PR #185, opened 2019-08-08
Use k3d instead of docker-compose

### LINKED ISSUE #179: Use k3d instead of compose to create integration cluster
k3d is simpler and faster than using compose to manage clusters.

dtomcej:
This PR:

- Installs k3d
- Uses k3d to create k3s cluster
- Remove image saving/mounting code

Fixes #179.


dtomcej:
Should also fix #212 in this PR at the same time.

iwilltry42:
We're working on a rewrite of k3d, which will also let you use it as a package. That should make it easier for you and more straightforward than using os.exec :)

dtomcej:
>We're working on a rewrite of k3d, which will also let you use it as a package. That should make it easier for you and more straightforward than using exec :)

Gods amongst men.

PR #210, opened 2019-09-06
Add support of coredns 1.5 and 1.6 

### LINKED ISSUE #207: Maesh  install  do not support  coredns 1.5.0
My croedns image is coredns/coredns:1.5.0
   When use helm install maesh,  maesh-prepare error log
`
time="2019-09-06T06:56:18Z" level=info msg="Checking Cluster..."
2019/09/06 06:56:18 command prepare error: error during cluster check: unsupported CoreDNS version "1.5.0", (supported versions are: 1.3,1.4)`

   In code  maesh/internal/k8s/client.go
   `var (
	supportedCoreDNSVersions = []string{
		"1.3",
		"1.4",
	}
   )`
   Is the code bug or really do not support ?

mmatur:
### Description

This PR adds support for CoreDNS 1.5 and 1.6

Fixes #207

### DIFF
--- a/.semaphoreci/setup.sh
+++ b/.semaphoreci/setup.sh
@@ -1,6 +1,7 @@
 # For personnal CI
 # mv /home/runner/workspace/src/github.com/<username>/ /home/runner/workspace/src/github.com/containous/
 # cd /home/runner/workspace/src/github.com/containous/maesh/
+for s in apache2 cassandra elasticsearch memcached mysql mongod postgresql sphinxsearch rethinkdb rabbitmq-server redis-server; do sudo service $s stop; done
 sudo swapoff -a
 sudo dd if=/dev/zero of=/swapfile bs=1M count=3072
 sudo mkswap /swapfile
--- a/integration/coredns_test.go
+++ b/integration/coredns_test.go
@@ -24,77 +24,52 @@ func (s *CoreDNSSuite) TearDownSuite(c *check.C) {
 	s.stopComposeProject()
 }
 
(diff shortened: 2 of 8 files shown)


.semaphoreci/setup.sh
```diff
@@ -1,6 +1,7 @@
 # For personnal CI
 # mv /home/runner/workspace/src/github.com/<username>/ /home/runner/workspace/src/github.com/containous/
 # cd /home/runner/workspace/src/github.com/containous/maesh/
+for s in apache2 cassandra elasticsearch memcached mysql mongod postgresql sphinxsearch rethinkdb rabbitmq-server redis-server; do sudo service $s stop; done
```
dtomcej: We are using the docker light box, where these are not present no?

integration/coredns_test.go
```diff
@@ -63,3 +64,37 @@ func (s *CoreDNSSuite) TestCoreDNSVersion140(c *check.C) {
 	s.waitKubectlExecCommand(c, argSlice, "whoami")
 	s.unInstallHelmMaesh(c)
 }
+
+func (s *CoreDNSSuite) TestCoreDNSVersion152(c *check.C) {
```
dtomcej: could these be abstracted to a single function `TestCoreDNSVersion(version string, c*check.C)`?
Since the only thing that seems to change is the string in the `setCoreDNSVersion` method?

.semaphoreci/setup.sh
mmatur: We are no more using docker light because there are too many flaky e2e tests with this type of boxes.

PR #215, opened 2019-09-10
Fix helm chart metrics version

mmatur:
### Description

This PR fix Namespace name in the metrics chart.

### DIFF
--- a/.semaphoreci/job1.sh
+++ b/.semaphoreci/job1.sh
@@ -2,4 +2,3 @@
 set -e
 
 make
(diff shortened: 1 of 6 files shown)


.semaphoreci/job1.sh
```diff
@@ -2,4 +2,3 @@
 set -e
 
 make
-if [ "$BRANCH_NAME" = "master" ]; then make push-docker ; fi
```
dtomcej: No more pushing docker?

mmatur: For now on the master branch no. We will only push stable version. 
For now it will be easier to manage

Issue #121, opened 2019-07-15
Re-evaluate helm


dtomcej:
Currently we use helm to install the mesh,

However, we may just be able to reduce the configuration to simple yaml, and `kubectl create -f`.

We should re-evaluate if we want to continue using helm for deployment, as it has its own issues and potential drawbacks.

Bessonov:
I'm very unhappy with helm as a deployment tool. But I'm very happy with helm's [helm template](https://helm.sh/docs/helm/#helm-template) and a some custom way to deploy the output.

Just as proposal and instead "helm vs. no helm" why not use helm, but generate a single `yaml` with `helm template` on CI?

dtomcej:
The major benefit to use helm is its hook system. Not having to write all the ordering logic into our installer scripts makes life much easier.

dtomcej:
The other benefit to using helm is that the chart becomes more accessible for systems that already use helm (like k3s etc), as the chart is easily deployed, as opposed to manually handling individual scripts.

george-angel:
I would love to see kustomize templates for deployment.

guibirow:
I would also love to see an option using Kustomize, or a plain old yaml 

Not everyone can use helm due to it's elevated privileges.

I've been working in quite a few projects and helm is either being re-evaluated or exchanged for Kustomize or plain yaml with minimal drawbacks, most people use it just because most tools documentation (like maesh) only provides helm as an installation option.

If removing helm from the quick start is not an option, at least should be provided a how to guide using yaml files.

Nginx ingress controller is a nice example on how to document the manual process.

Bessonov:
> Not everyone can use helm due to it's elevated privileges.

That's not true. I use helm without tiller and elevated privileges.



> I would love to see kustomize templates for deployment.

From my point of view, kustomize is an another workaround like helm, but not a solution. Both are very usable for simple use cases, but if it comes to deeper customisation like removing of resources or iterating, they can't still help there. Or I'm missing something?

carlosedp:
Another option would be converting the manifests to Jsonnet with https://github.com/ksonnet/ksonnet-lib mixins. It would reduce the amount of written code and still have flexibility.

dtomcej:
@carlosedp, we are currently re-evaluating our hooks in #252, in an effort to reduce down to plain yaml for installation.

This would most likely simplify a lot of options, as `helm template` would generate all the yaml needed to install.

PR #216, opened 2019-09-10
Add non-helm quickstart

nbyl:
While helm, of which I am very fan, is a great and flexible for deploying kubernetes applications, it also creates a barrier. This PR creates a yaml manifest that can be installed with a single commandline without prerequisites. I hope this will maesh more accessible for users, as discussed in #121.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ deploy:
       condition: $STABLE = true
   - provider: script
     skip_cleanup: true
-    script: make helm-package
+    script: make helm-package helm-manifest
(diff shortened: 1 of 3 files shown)


Bessonov:
@nbyl I'm not sure, that adding namespace should be default.

@SantoDE Thank you )

dtomcej:
Hello @nbyl,

Since we haven't had any activity on this PR, we are going to close it.

Please feel free to re-open another PR if you feel that Maesh does not meet your needs.

Thanks!

PR #230, opened 2019-09-16
add static configuration option for ignore namespaces

### LINKED ISSUE #220: Improve Configuration Refresh logic
#### Proposal

There are 2 different areas which I would like to tackle in terms of configuration refresh.

a) There are Objects which "spamm" the Kubernetes API and therefore will force an ongoing Maesh Configuration refresh. There should be an option to ignore certain namespaces / services
b) If the Deploy queue is filled, we should check if there is an older "entry" for that maesh pod already. If so, remove it

 #### Background

Make the process more resilient and faster

 #### Workarounds

SantoDE:
This one fixes #220 :)

There is know a new option to ignore "namespaces"

### DIFF
--- a/helm/chart/maesh/values.yaml
+++ b/helm/chart/maesh/values.yaml
@@ -16,6 +16,7 @@ controller:
       mem: "50Mi"
       cpu: "100m"
   logging:
+    excludeNamespaces:
     debug: true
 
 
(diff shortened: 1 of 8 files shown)


helm/chart/maesh/values.yaml
```diff
@@ -16,6 +16,7 @@ controller:
       mem: "50Mi"
       cpu: "100m"
   logging:
+    excludeNamespaces:
```
mmatur: It should not be under `controller.logging`

SantoDE: Moved

Issue #225, opened 2019-09-12
Installation issue


huang195:
I'm simply trying to kick the tires and followed the installation instructions here: https://docs.mae.sh/install/

`helm repo` commands worked fine, but the last step to install the helm charts crapped out
```
# helm install --name=maesh --namespace=maesh maesh/maesh
Error: serviceaccounts "maesh-prepare" is forbidden: User "system:serviceaccount:kube-system:default" cannot delete resource "serviceaccounts" in API group "" in the namespace "maesh"
```

I'm suspecting I missed a prereq step? I'm running in a vanilla K8s 1.14 cluster.

ereslibre:
You can fix this by adding the following:

```
- kind: ServiceAccount                                                               
  name: default                                                                      
  namespace: kube-system
```

on the `subjects` section in the `cluster-admin` `ClusterRoleBinding`.

jsPlachetka:
@ereslibre Thank you! I had the same problem.

Would it make sense to add it to the quickstart documentation?

SantoDE:
It would totally make sense :) Are you willing to submit a PR? :)

ereslibre:
Please note that this is a solution for kicking the tires in a fast way, probably not meant for production setups :)

Also, as far as I can see, the team is reconsidering Helm (https://github.com/containous/maesh/issues/121), so depending on the chosen solution this would not be necessary.

SantoDE:
I agree, but still we could at least add a note to the docs for that :) So even when we drop helm, until that we will have somethings on the docs

PR #233, opened 2019-09-16
Add note about tiller RBAC on the quickstart guide

ereslibre:
Add instructions on how to configure the `default` ServiceAccount on
the `kube-system` namespace so it acts as a `cluster-admin`.

While not recommended for production, it's good enough to give maesh a
try on an isolated environment.

Fixes: https://github.com/containous/maesh/issues/225


mmatur:
Thanks for your contribution @ereslibre

PR #239, opened 2019-09-17
Use k3d to create k8s cluster instead of compose

dtomcej:
By using k3d, we have more control, and more logging than we do with compose.

We also have a cleaner interface to load in images, rather than mounting tarballs.


dtomcej:
>what do we wanna do with the "old" test-integration target in the Makefile

I think we should remove it, as we are unlikely going to go back to testing a single binary.

PR #246, opened 2019-09-19
Add support for kubedns

mmatur:
### Description

This PR adds support for kubedns.

If KubeDNS is detected in the kubernetes cluster and coredns is not installed. Maesh will install CoreDNS as a DaemonSet and will patch KubeDNS to add a stubDomains that forward all dns request with prefix `.maesh` to the coreDNS instances.

### DIFF
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -33,5 +33,5 @@ Using the Maesh endpoints is all that is required.
 To run this app, you require the following:
 
 - Kubernetes 1.11+
-- CoreDNS installed as [Cluster DNS Provider](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/) (versions 1.3+ supported)
+- CoreDNS/KubeDNS installed as [Cluster DNS Provider](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/) (versions 1.3+ supported)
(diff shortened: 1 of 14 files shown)


mmatur:
Yes it could be a better idea to have a Deployment with anti affinity. By default k3s deploy only replica of coredns.

nbyl:
From an end-user perspective: Having more than DNS instance is always a good idea. But we usually don't deploy one instance per machine. So having this as deployment with configurable replica count would be nice.

SantoDE:
I agree @nbyl,

that would be pretty cool. However, it will require to rethink the current helm values.

What do you think of something like

```
core-dns:
    deploy: true
    replicas: 2
```
instead of `kubednes: true`? Any other suggestion?

donnyv12:
I think since Maesh gets deployed as a daemonset, having CoreDNS as a daemonset in KubeDNS clusters makes sense. Also, it sounds like CoreDNS will only handle .maesh requests, so KubeDNS will continue to function the same it is prior to installing Maesh. Correct me if I am wrong there. I feel like this is a great step towards GKE support, which I am very excited about.

nesl247:
Please do not depend on helm hook support. This will make this incredibly difficult to use with tools like Pulumi which only use helm templating.

PR #247, opened 2019-09-19
Make Ignorenamespaces better manageable

SantoDE:
This PR makes it possible, to manage multiple namespaces to ignore in a more "listable" Style

```controller:
  ignoreNamespaces: 
    - local-path-storage
    - some-other
```

### DIFF
--- a/helm/chart/maesh/templates/_helpers.tpl
+++ b/helm/chart/maesh/templates/_helpers.tpl
@@ -14,7 +14,7 @@ Define the templated image with tag
     {{- printf "%s:%s" .Values.controller.image.name ( .Values.controller.image.tag | default .Chart.AppVersion ) -}}
 {{- end -}}
 {{/*
-Define the templated image with tag
+Define the ignoreNamespaces List
 */}}
 {{- define "maesh.controllerIgnoreNamespaces" -}}
     --ignoreNamespaces=
--- a/helm/chart/maesh/templates/_helpers.tpl
+++ b/helm/chart/maesh/templates/_helpers.tpl
@@ -7,10 +7,19 @@ Define the Chart version Label
 {{- define "maesh.chartLabel" -}}
     {{- printf "%s-%s" .Chart.Name .Chart.Version -}}
 {{- end -}}
-
 {{/*
 Define the templated image with tag
 */}}
 {{- define "maesh.controllerImage" -}}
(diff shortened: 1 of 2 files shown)


helm/chart/maesh/templates/_helpers.tpl
```diff
@@ -7,10 +7,19 @@ Define the Chart version Label
...
 {{- end -}}
-
 {{/*
 Define the templated image with tag
 */}}
 {{- define "maesh.controllerImage" -}}
     {{- printf "%s:%s" .Values.controller.image.name ( .Values.controller.image.tag | default .Chart.AppVersion ) -}}
 {{- end -}}
+{{/*
+Define the templated image with tag
```
dtomcej: Can we get a better comment here 😺

helm/chart/maesh/templates/_helpers.tpl
```diff
@@ -7,10 +7,19 @@ Define the Chart version Label
...
+*/}}
+{{- define "maesh.controllerIgnoreNamespaces" -}}
+    --ignoreNamespaces=
+    {{- range $idx, $ns := .Values.controller.ignoreNamespaces }}
+        {{- if $idx }},{{ end }}
+            {{- $ns }}
+    {{- end -}}
+{{- end -}}
```
dtomcej: newline

SantoDE:
Done @dtomcej :)

helm/chart/maesh/templates/_helpers.tpl
```diff
@@ -7,10 +7,19 @@ Define the Chart version Label
...
 */}}
 {{- define "maesh.controllerImage" -}}
     {{- printf "%s:%s" .Values.controller.image.name ( .Values.controller.image.tag | default .Chart.AppVersion ) -}}
 {{- end -}}
+{{/*
+Define the ignoreNamespaces List
+*/}}
+{{- define "maesh.controllerIgnoreNamespaces" -}}
+    --ignoreNamespaces=
+    {{- range $idx, $ns := .Values.controller.ignoreNamespaces }}
```
dtomcej: What happens if this is empty?
does it just result in:
```
--ignoreNamespaces=
```
?

And does that break anything?

SantoDE: Yes it will create an empty --ignoreNamespaces, but it doesn't break anything. However, I added a sanity check. Thanks!

Issue #242, opened 2019-09-18
Move Maesh to use stateless providers


dtomcej:
#### Proposal

Keeping state requires ordering, and maintenance between messages.

Moving to stateless providers would remove the requirement for state maintenance, and would guarantee atomic configuration updates.

 #### Background

Maesh currently maintains state, and does not guarantee atomic configuration updates.

This causes Queue backlogs such as #221, and #219 

Going stateless in combination with #241, should **GREATLY** reduce the processing time for configuration updates

#### Requirements

This issue requires:
- [x] Migrating Kubernetes Provider - #251 
- [x] Migrating SMI Provider - #260 
- [ ] Removing state operations from Controller

PR #251, opened 2019-09-24
Move to stateless Kubernetes Provider

dtomcej:
This PR:

- modifies the Kubernetes provider to be stateless.
- reduces the number of kubernetes api calls by selecting objects first, then looping through them in-memory
- moves base config building to the provider, from the controller.

Partially solves #242.

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -379,8 +379,7 @@ func (c *Controller) createMeshServices() error {
 			}
 		}
 		log.Infof("Creating associated mesh service: %s", meshServiceName)
-		err := c.createMeshService(service)
-		if err != nil {
+		if err := c.createMeshService(service); err != nil {
 			return fmt.Errorf("unable to get create mesh service: %v", err)
 		}
 	}
@@ -417,6 +416,7 @@ func (c *Controller) createMeshService(service *corev1.Service) error {
 
 			if targetPort.IntVal == 0 {
 				log.Errorf("Could not get TCP Port for service: %s with service port: %v", service.Name, sp)
+				continue
 			}
 
 			meshPort := corev1.ServicePort{
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -182,8 +182,9 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 	}
 
 	// Create the mesh services here to ensure that they exist
+	log.Info("Creating initial mesh services")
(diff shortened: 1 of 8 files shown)


dtomcej:
Build is running into issues due to the fact that the k8s provider tries to pull all data before the mesh services have a chance to be built.

This causes discrepancies due to the TCP state table not being populated until the mesh service is created.

dtomcej:
We also need to move the Mesh service creation to also be stateless.

CRUD operations need to be atomic as well, so that they can be processed before the provider returns the atomic config.

internal/controller/controller.go
```diff
@@ -362,11 +360,39 @@ func (c *Controller) processDeletedMessage(event message.Message) {
...
+		meshServiceName := c.userServiceToMeshServiceName(service.Name, service.Namespace)
+		for _, subservice := range services {
+			// If there is already a mesh service created, don't bother recreating
+			if subservice.Name == meshServiceName && subservice.Namespace == c.meshNamespace {
+				continue
+			}
+		}
+		log.Infof("Creating associated mesh service: %s", meshServiceName)
+		err := c.createMeshService(service)
+		if err != nil {
```
jlevesy: WDYT of inilining this expression

```
if err := c.createMeshService(service); err != nil {
```

internal/controller/controller.go
```diff
@@ -389,6 +415,10 @@ func (c *Controller) createMeshService(service *corev1.Service) (*corev1.Service
 				targetPort = intstr.FromInt(c.getTCPPortFromState(service.Name, service.Namespace, sp.Port))
 			}
 
+			if targetPort.IntVal == 0 {
+				log.Errorf("Could not get TCP Port for service: %s with service port: %v", service.Name, sp)
```
jlevesy: Shoudn't we skip this item in that case using a `continue` statement ?

internal/k8s/client.go
```diff
@@ -585,6 +587,20 @@ func (w *ClientWrapper) GetEndpoints(namespace, name string) (*corev1.Endpoints,
 	return endpoints, exists, err
 }
 
+// GetEndpointses retrieves the endpoints from all namespaces.
```
jlevesy:
```suggestion
// GetEndpointses retrieves the endpoints from given namespace.
```

internal/controller/controller.go
dtomcej: Good catch!

PR #254, opened 2019-09-25
Move hooks to init container and subdir

### LINKED ISSUE #252: Review Hooks for necessity
#### Proposal

We should review hooks (such as prepare and coreDNS) and determine if they could be moved to init containers, or jobs instead of relying on helm for timing and completion.

 #### Background

 Lessening our reliance on helm makes maintenance easier, and gives users more options for deployment/installation methods.

dtomcej:
This PR:

- Removes and moves hooks to init containers and chart subdirs
- Removes the init container SA, since the perms are almost identical

Works towards #121.

Fixes #252


dtomcej:
Tests are failing, because the job I removed can't fail ;)

dtomcej:
This PR also fixes some lint issues found on golangci-lint 1.19.0

PR #255, opened 2019-09-30
maesh-prepare-hook as non-root

bavarianbidi:


dtomcej:
Should be blocked by #254 

So that we can have the controller run as non-root as well.

dtomcej:
Hello @bavarianbidi,

Thanks for your patience with this. We would love you to update this PR now that we have completed rewriting the hooks.

Thanks!

bavarianbidi:
PR could be closed, because `prepare` is now an initContainer in the `controller-deployment` and there the securityContext-values are set correctly

PR #260, opened 2019-10-01
Move SMI provider to be stateless

dtomcej:
This PR:

- Updates the SMI provider to be stateless

Todo:

- [x] Update unit tests to handle new provider behavior
- [x] Update build config to request objects as a list instead of requesting objects in loops
- [x] Move duplicated functions like `getEndpointsFromList` and `createBaseConfigWithReadiness` to a package so they are not duplicated

Partially solves #242.


SantoDE:
Build fails on semaphore

PR #266, opened 2019-10-04
Check for Jaeger values set or take default

### LINKED ISSUE #265: Don't see any trace on Jaeger
Hello all, 

I want to play with Maesh, and give a talk about it in my company, but i'm not able to see any trace in jaeger UI, other than jaeger internals.

My demo application if a simple front, calling a backend-api (via http://backend.back.maesh) 

Am i doing something wrong ? 

Seems related to #240

SantoDE:
Partially fixes #265 

This will allow the correct default values to be set

### DIFF
--- a/helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
+++ b/helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
@@ -67,12 +67,12 @@ spec:
               {{- if .Values.tracing.jaeger.localagenthostport }}
             - "--tracing.jaeger.localagenthostport={{ .Values.tracing.jaeger.localagenthostport }}"
               {{- else }}
-            - "--tracing.jaeger.localagenthostport=jaeger-agent.maesh.svc.cluster.local:6831"
+            - "--tracing.jaeger.localagenthostport=jaeger-agent.{{ .Release.Namespace }}.svc.cluster.local:6831"
               {{- end }}
               {{- if .Values.tracing.jaeger.localagenthostport }}
             - "--tracing.jaeger.samplingserverurl={{ .Values.tracing.jaeger.samplingserverurl }}"
               {{- else }}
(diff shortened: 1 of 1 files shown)


dtomcej:
I would suggest `{{ default "defaultvalue" .Value }}` to use a value if not set:

helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
```diff
@@ -64,8 +64,16 @@ spec:
             - "--providers.rest"
             - "--providers.rest.insecure"
             {{- if .Values.tracing.jaeger.enabled }}
+              {{- if .Values.tracing.jaeger.localagenthostport }}
             - "--tracing.jaeger.localagenthostport={{ .Values.tracing.jaeger.localagenthostport }}"
+              {{- else }}
+            - "--tracing.jaeger.localagenthostport=jaeger-agent.maesh.svc.cluster.local:6831"
```
ADefossez-Zenika: Hi, IMHO if you install maesh in another namespace than the default one, you should replace this line with something like : 
            - "--tracing.jaeger.localagenthostport=jaeger-agent.{{ .Release.Namespace }}.svc.cluster.local:6831"

helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
```diff
@@ -64,8 +64,16 @@ spec:
...
+              {{- end }}
+              {{- if .Values.tracing.jaeger.localagenthostport }}
             - "--tracing.jaeger.samplingserverurl={{ .Values.tracing.jaeger.samplingserverurl }}"
+              {{- else }}
+            - "--tracing.jaeger.samplingserverurl=http://jaeger-agent.maesh.svc.cluster.local:5778/sampling"
```
ADefossez-Zenika: same thing here

PR #270, opened 2019-10-08
Remove Queues, and use channels for processing

### LINKED ISSUE #219: Refresh Configuration when SMI Crd is deleted
#### Proposal

Rebuild the maesh configuration once SMI spec CRD's are deleted.

 #### Background

Change in the SMI CRD's might result in a service not being allowed to take access another service anymore. Therefore the configuration should refresh.

 #### Workarounds

Restart Maesh Controller pod

### LINKED ISSUE #221: Remove older items from deploy queue when new item is added
#### Proposal

Older items (per pod) should be removed from the deploy queue when new items are added (per pod)

 #### Background

 There is no need to keep older versions of code deploys in the deploy queue when we have a newer version in the queue.

### LINKED ISSUE #241: Don't redeploy a configuration if the configuration has not changed
#### Proposal

If the configuration has not changed, don't deploy a new copy of the same

 #### Background

 If k8s api updates trigger a re-deploy, don't deploy a new version with the same actual configuration.

This will prevent useless config deployments.

### LINKED ISSUE #243: Remove Processing Queues
#### Proposal

Maesh currently uses queues for data processing. We want to move to channels, or another option to reduce processing/waiting overhead, and manage the data immediately.

 #### Background

 Queues are great, but in some instances, old data is not usable as soon as a newer version is created. This causes issues when you have old configurations in queues while new configurations are being created.

This combined with non-atomic updates causes latency and processing overhead when it is not needed.

### LINKED ISSUE #263: Create Provider Interface
#### Proposal

Register `BuildConfig` handler with the controller so that the if smi is not required after controller startup

Also create an interface so that the controller can interact with the interface, and not the specific providers

 #### Background

Cleaner code, and will work better with future providers

dtomcej:
This PR
- Removes Queues, and uses channels for triggers
- Creates a base provider interface
- Registers provider functions to eliminate redundancies
- Creates a `lastConfiguration` to prevent duplicate configuration deploys

Fixes #219 
Fixes #221 
Fixes #241 
Fixes #243 
Fixes #263


dtomcej:
@mmatur it appears that this is failing specifically with the trafficsplit integration test.

Do you have any suggestions as to why it would be blowing up?

dtomcej:
I think it has to do with the `createMeshService` not getting triggered 😄 

Will look into that tomorrow morning 😄

PR #273, opened 2019-10-09
Add SMI Support for TCP

### LINKED ISSUE #264: Ensure TCP support via SMI provider
#### Proposal

 We need to ensure that TCP service support is enabled via the SMI provider

 #### Background

 We have encountered issues in the past with TCP and the SMI provider

SantoDE:
Fixes #264


internal/providers/smi/fixtures/mock_tcp.yaml
```diff
@@ -0,0 +1,20 @@
...
+    name: prometheus
+    namespace: default
+specs:
+  - kind: TCPRoute
+    name: api-service-routes
+---
+kind: TCPRoute
+apiVersion: specs.smi-spec.io/v1alpha1
+metadata:
+  name: api-service-routes
```
dtomcej: newline

integration/resources/smi/access-control/5.trs.yaml
```diff
@@ -0,0 +1,5 @@
+---
+apiVersion: specs.smi-spec.io/v1alpha1
+kind: TCPRoute
+metadata:
+  name: tcp-access
```
dtomcej: newline

PR #275, opened 2019-10-09
Feature/scheme service

mmatur:
### Description

Allow user to customize scheme.

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -32,8 +32,15 @@ The service mode can be enabled by using the following annotation:
 maesh.containo.us/traffic-type: "http"
 ```
 
-This annotation can be set to either `http` or `tcp`, and will specify the mode for that service operation.
-If this annotation is not present, the mesh service will operate in the default mode specified in the static configuration.
+### Scheme
+
+The scheme used to define custom scheme for request:
+
+```yaml
+maesh.containo.us/scheme: "h2c"
+```
+
+This annotation can be set to either `http` or `h2c`.
 
 ### Retry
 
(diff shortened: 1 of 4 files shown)


docs/content/configuration.md
```diff
@@ -32,8 +32,15 @@ The service mode can be enabled by using the following annotation:
 maesh.containo.us/traffic-type: "http"
 ```
 
-This annotation can be set to either `http` or `tcp`, and will specify the mode for that service operation.
-If this annotation is not present, the mesh service will operate in the default mode specified in the static configuration.
+### Scheme
```
dtomcej: Did we want to remove the description for the previous annotation?

Also, we should probably mention that scheme is only applicable if `traffic-type` is http

internal/providers/kubernetes/provider.go
```diff
@@ -180,6 +181,16 @@ func (p *Provider) getServiceMode(annotations map[string]string) string {
 	return mode
 }
 
+func (p *Provider) getScheme(annotations map[string]string) string {
+	scheme := annotations[k8s.AnnotationScheme]
+
+	if scheme == "" {
```
dtomcej: we should validate that the scheme is an acceptable scheme here.

internal/providers/kubernetes/provider.go
```diff
@@ -180,6 +181,16 @@ func (p *Provider) getServiceMode(annotations map[string]string) string {
 	return mode
 }
 
+func (p *Provider) getScheme(annotations map[string]string) string {
```
dtomcej: since this doesn't actually use the provider at all, did we want to have this method moved to the `base` package so that we can use it with the SMI provider?

dtomcej:
Did we want to add this functionality to the SMI provider in this PR? or did we want to add it in another?

PR #283, opened 2019-10-15
make cluster domain configurable

### LINKED ISSUE #278: Helm chart is assuming a `cluster.local` domain
This is not true for our clusters

```
☸ this ☁  maesh [master] ag cluster.local
internal/k8s/client.go
365:        name regex ([a-zA-Z0-9-_]*)\.([a-zv0-9-_]*)\.maesh maesh-{1}-{2}.maesh.svc.cluster.local
368:    kubernetes cluster.local in-addr.arpa ip6.arpa {

integration/kubedns_test.go
31:		"exec", "-it", pod.Name, "-n", pod.Namespace, "-c", pod.Spec.Containers[0].Name, "--", "curl", "whoami.whoami.svc.cluster.local", "--max-time", "5",

integration/resources/kubedns/kubedns.yaml
122:            - --domain=cluster.local.
163:            - --server=/cluster.local/127.0.0.1#10053
195:            - --probe=kubedns,127.0.0.1:10053,kubernetes.default.svc.cluster.local,5,SRV
196:            - --probe=dnsmasq,127.0.0.1:53,kubernetes.default.svc.cluster.local,5,SRV

integration/resources/coredns/coredns.yaml
58:        kubernetes cluster.local in-addr.arpa ip6.arpa {
```

brennerm:
Fixes #278 

- add new variable `clusterDomain` to chart
- use variable instead of hard coded `cluster.domain`

### DIFF
--- a/helm/chart/maesh/charts/metrics/templates/grafana.yaml
+++ b/helm/chart/maesh/charts/metrics/templates/grafana.yaml
@@ -117,7 +117,7 @@ data:
     - name: prometheus
       type: prometheus
       orgId: 1
-      url: http://prometheus.{{ .Release.Namespace }}.svc.cluster.local:9090
+      url: http://prometheus.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:9090
       isDefault: true
       access: proxy
       jsonData:
--- a/helm/chart/maesh/charts/metrics/templates/grafana.yaml
+++ b/helm/chart/maesh/charts/metrics/templates/grafana.yaml
@@ -117,7 +117,7 @@ data:
     - name: prometheus
       type: prometheus
       orgId: 1
-      url: http://prometheus.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:9090
+      url: http://prometheus.{{ .Release.Namespace }}.svc.{{ default "cluster.local" .Values.clusterDomain }}:9090
       isDefault: true
       access: proxy
       jsonData:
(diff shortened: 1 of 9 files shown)


dtomcej:
This is close, but there is a raw portion of the configmap that is patched via the `prepare` command:

https://github.com/containous/maesh/blob/master/internal/k8s/client.go#L360

This means that the prepare command needs to have this passed as a flag, and then passed onto the k8s client for patching.

helm/chart/maesh/values.yaml
```diff
@@ -24,6 +24,7 @@ controller:
   affinity: {}
 
 kubedns: false
+clusterDomain: cluster.local
```
dtomcej: Should be commented out by default

helm/chart/maesh/charts/metrics/templates/grafana.yaml
```diff
@@ -117,7 +117,7 @@ data:
     - name: prometheus
       type: prometheus
       orgId: 1
-      url: http://prometheus.{{ .Release.Namespace }}.svc.cluster.local:9090
+      url: http://prometheus.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:9090
```
dtomcej: Should use the default here, so that it doesn't have to be set manually:

```
{{ default "cluster.local" .Values.clusterDomain }}
```

dtomcej:
Hi @brennerm,

This PR looks great, but we need you to sign the CLA to allow us to move forward with review.

Could you do so?

Thanks!

brennerm:
Done :+1:

PR #284, opened 2019-10-16
Remove unused example variable

frelon:
Tag is not used in `mesh-daemonset.yaml`.


dtomcej:
@mmatur @SantoDE Do you think that we should remove this tag?

Or do you think that we should add a templated image/tag like we do for the controller?

mmatur:
Personally, I prefer the option where we have a templated image/tag like we do for the controller

Issue #244, opened 2019-09-18
Parallel configuration deployments


dtomcej:
#### Proposal

We should be deploying configurations to mesh nodes in parallel, so that nodes are non-blocking. We could use an EBO, since the deploy would be non-blocking for other nodes.

 #### Background

 Currently node deployment is sequential, so old configurations get retried, and block new ones.

dtomcej:
Since moving to having no queues to block, this can be done with a single fail, and the audit timer should catch and retry.

This means that we could use a wait group and deploy in parallel.

jlevesy:
I'll work on this one

PR #337, opened 2019-11-06
Deploy to pods in parallel

jlevesy:
### What does this PR do

This PR makes the maesh controller deploy dynamic configuration in parallel.

Fixes: #244

### Additional information

- I'm relying on `golang.org/x/sync/errgroup` which is equivalent to `sync.WaitGroup` but cleaner. Is it OK ? 
- Each deployment goroutine applies a backoff with an hardcoded max elapsed time of ~1 minute~ 15 seconds. ~Do we want to make this configurable ?~
- I unattached `deployToPod` from  the Controller as it does not depend on any members of this struct. WDYT? 
- ~Opened as a draft 'cuz I'm not quite done testing :sweat_smile:~

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -511,7 +511,7 @@ func (c *Controller) deployConfiguration(config *dynamic.Configuration) error {
 		return fmt.Errorf("unable to find any active mesh pods to deploy config : %+v", config)
 	}
 
-	if err := deployToPods(podList.Items, config); err != nil {
+	if err := c.deployToPods(podList.Items, config); err != nil {
 		return fmt.Errorf("error deploying configuration: %v", err)
 	}
 
@@ -542,14 +542,14 @@ func (c *Controller) deployConfigurationToUnreadyNodes(config *dynamic.Configura
 		}
 	}
 
-	if err := deployToPods(unreadyPods, config); err != nil {
+	if err := c.deployToPods(unreadyPods, config); err != nil {
 		return fmt.Errorf("error deploying configuration: %v", err)
 	}
 
 	return nil
 }
 
-func deployToPods(pods []corev1.Pod, config *dynamic.Configuration) error {
+func (c *Controller) deployToPods(pods []corev1.Pod, config *dynamic.Configuration) error {
 	var errg errgroup.Group
 
 	for _, p := range pods {
@@ -562,7 +562,7 @@ func deployToPods(pods []corev1.Pod, config *dynamic.Configuration) error {
 			b.MaxElapsedTime = 15 * time.Second
 
 			op := func() error {
-				return deployToPod(pod.Name, pod.Status.PodIP, config)
+				return c.deployToPod(pod.Name, pod.Status.PodIP, config)
 			}
 
 			return backoff.Retry(safe.OperationWithRecover(op), b)
@@ -572,7 +572,7 @@ func deployToPods(pods []corev1.Pod, config *dynamic.Configuration) error {
 	return errg.Wait()
 }
 
-func deployToPod(name, ip string, config *dynamic.Configuration) error {
+func (c *Controller) deployToPod(name, ip string, config *dynamic.Configuration) error {
 	if name == "" || ip == "" {
 		// If there is no name or ip, then just return.
 		return fmt.Errorf("pod has no name or IP")
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -559,7 +559,7 @@ func deployToPods(pods []corev1.Pod, config *dynamic.Configuration) error {
 
 		errg.Go(func() error {
 			b := backoff.NewExponentialBackOff()
-			b.MaxElapsedTime = time.Minute
+			b.MaxElapsedTime = 15 * time.Second
 
 			op := func() error {
 				return deployToPod(pod.Name, pod.Status.PodIP, config)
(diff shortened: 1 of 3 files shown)


internal/controller/controller.go
```diff
@@ -507,18 +509,68 @@ func (c *Controller) deployConfiguration(config *dynamic.Configuration) error {
...
+	if err != nil {
+		return fmt.Errorf("unable to retrieve pod list: %v", err)
+	}
+
+	if len(podList.Items) == 0 {
+		return fmt.Errorf("unable to find any active mesh pods to deploy config : %+v", config)
+	}
+
+	var unreadyPods []corev1.Pod
+
```
Ullaakut:
```suggestion
```

internal/controller/controller.go
```diff
@@ -507,18 +509,68 @@ func (c *Controller) deployConfiguration(config *dynamic.Configuration) error {
...
 	return nil
 }
 
-func (c *Controller) deployToPod(name, ip string, config *dynamic.Configuration) error {
+func deployToPods(pods []corev1.Pod, config *dynamic.Configuration) error {
+	var errg errgroup.Group
+
+	for _, p := range pods {
+		pod := p
```
dtomcej: Why this assignment?

internal/controller/controller.go
```diff
@@ -507,18 +509,68 @@ func (c *Controller) deployConfiguration(config *dynamic.Configuration) error {
...
+
+		log.Debugf("Deploying to pod %s with IP %s", pod.Name, pod.Status.PodIP)
+
+		errg.Go(func() error {
+			b := backoff.NewExponentialBackOff()
+			b.MaxElapsedTime = time.Minute
```
dtomcej: Did we want to fail after a minute? Perhaps 30 seconds might be a saner timeout.

internal/controller/controller.go
```diff
@@ -507,18 +509,68 @@ func (c *Controller) deployConfiguration(config *dynamic.Configuration) error {
...
+	if len(podList.Items) == 0 {
+		return fmt.Errorf("unable to find any active mesh pods to deploy config : %+v", config)
+	}
+
+	var unreadyPods []corev1.Pod
+
+	for _, pod := range podList.Items {
+		for _, status := range pod.Status.ContainerStatuses {
+			if !status.Ready {
+				unreadyPods = append(unreadyPods, pod)
```
dtomcej: If there are two unready containers in the pod, won't this result in the pod being appended twice?

internal/controller/controller.go
dtomcej: I'm ok with a quicker fail, since we have a 10s timer re-triggering a deploy to non-ready nodes

internal/controller/controller.go
```diff
@@ -507,18 +509,68 @@ func (c *Controller) deployConfiguration(config *dynamic.Configuration) error {
...
 
+	if err := deployToPods(unreadyPods, config); err != nil {
+		return fmt.Errorf("error deploying configuration: %v", err)
+	}
+
 	return nil
 }
 
-func (c *Controller) deployToPod(name, ip string, config *dynamic.Configuration) error {
+func deployToPods(pods []corev1.Pod, config *dynamic.Configuration) error {
```
dtomcej: This unattachment may break with the new deploylog, although we could pass the log in through parameters.

Not sure which is preferrable

internal/controller/controller.go
SantoDE: quicker is better, I think as well 👍

internal/controller/controller.go
SantoDE: yeah, true.

internal/controller/controller.go
jlevesy: Sound good to me, will do !

internal/controller/controller.go
jlevesy: Nice catch, thanks !

jlevesy: Actually no this code is correct. Once it hits the `break` statement it will break from the status loop and iterate to the next pod.

internal/controller/controller.go
jlevesy: This makes the linter complain :/

internal/controller/controller.go
jlevesy: `p` is a variable shared across mutliple goroutines by closure.

However, the p variable is assigned at each iteration. Which means that when the goroutines accessing the variable will start to execute and iterate over them, they will only get the last assigned value, in other words the last item of the slice.

Here is a simple example demonstrating the problem https://play.golang.org/p/fpmNrJ4MMNn

internal/controller/controller.go
SantoDE: True, I didnt see the break! nice

internal/controller/controller.go
Ullaakut: The linter is wrong here, it doesn't seem to understand that the unreadyPods is used in the nested for loop :( I'd suggest disabling this linter module or at least adding an exception for this because here I find this newline confusing :p

dtomcej: I like having the newline, and agree with the linter. Although the var is used in the loop, it is not used in the iteration.

dtomcej:
Can this be rebased? We need to figure out how to handle the DeployLog changes in this PR.

jlevesy:
Will do sir.

internal/controller/controller.go
jlevesy: I reattached, we will deal with this later.

PR #342, opened 2019-11-06
Create a cap for the max number of deploy log entries

### LINKED ISSUE #334: Create cap for deployLog size
#### Proposal

We should not allow the deploylog to grow indefinitely, as it could get large in large busy clusters. We should have a `maxRecords`, and drop old records when new ones are created.

dtomcej:
This PR:

- Creates a log rotation scheme where old records are removed
- Adds tests to ensure the rotation works

Fixes #334

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -99,7 +99,7 @@ func (c *Controller) Init() error {
 
 	c.tcpStateTable = &k8s.State{Table: make(map[int]*k8s.ServiceWithPort)}
 
-	c.deployLog = NewDeployLog()
+	c.deployLog = NewDeployLog(1000)
 	c.api = NewAPI(c.apiPort, &c.lastConfiguration, c.deployLog, c.clients, c.meshNamespace)
 
 	if c.smiEnabled {
--- a/internal/controller/log.go
+++ b/internal/controller/log.go
@@ -63,8 +63,3 @@ func (d *DeployLog) LogDeploy(timeStamp time.Time, podName string, podIP string,
 func (d *DeployLog) GetLog() []Entry {
 	return d.entries
 }
-
-// Len returns the number of records in the entries slice.
-func (d *DeployLog) Len() int {
-	return len(d.entries)
-}
--- a/internal/controller/log.go
+++ b/internal/controller/log.go
@@ -1,13 +1,13 @@
 package controller
 
 import (
-	"encoding/json"
 	"time"
 
 	log "github.com/sirupsen/logrus"
 )
 
-type logEntry struct {
+// Entry holds the details of a deployment.
+type Entry struct {
 	TimeStamp        time.Time
 	PodName          string
 	PodIP            string
@@ -17,7 +17,7 @@ type logEntry struct {
 
 // DeployLog holds a slice of log entries.
 type DeployLog struct {
-	entries    []logEntry
+	entries    []Entry
 	maxEntries int
 }
 
@@ -43,7 +43,7 @@ func (d *DeployLog) Init() error {
 
 // LogDeploy adds a record to the entries list.
 func (d *DeployLog) LogDeploy(timeStamp time.Time, podName string, podIP string, deploySuccessful bool, reason string) {
-	newEntry := logEntry{
+	newEntry := Entry{
 		TimeStamp:        timeStamp,
 		PodName:          podName,
 		PodIP:            podIP,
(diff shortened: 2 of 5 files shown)


internal/controller/log.go
```diff
@@ -48,15 +51,25 @@ func (d *DeployLog) LogDeploy(timeStamp time.Time, podName string, podIP string,
...
+	data, err := json.Marshal(d.entries)
 	if err != nil {
 		log.Error("Could not marshal deploylog entries")
 	}
 
 	return data
 }
+
+// GetLogLength returns the number of records in the entries slice.
+func (d *DeployLog) GetLogLength() int {
```
jlevesy: This would probably mean a `deployLog.GetLogLenth()`, WDYT of this instead ?
```suggestion
func (d *DeployLog) Len() int {
```

internal/controller/log.go
```diff
@@ -48,15 +51,25 @@ func (d *DeployLog) LogDeploy(timeStamp time.Time, podName string, podIP string,
...
+	if len(d.entries) >= d.maxEntries {
+		// Pull elements off the front of the slice to make sure that the newly appended record is under the max record value.
+		d.entries = d.entries[(len(d.entries)-d.maxEntries)+1:]
+	}
+
+	d.entries = append(d.entries, newEntry)
 }
 
 // GetLog returns a json representation of the entries list.
 func (d *DeployLog) GetLog() []byte {
```
jlevesy: That's more a general remark. But I would remove the serialization logic from here and move it to the API handler to return a  `[]Entry` here.

Only the API needs this data as JSON, and I don't like having API logic in this common object that one could want to reuse at some point.  This reduces reusability.

internal/controller/controller.go
```diff
@@ -97,7 +97,7 @@ func (c *Controller) Init() error {
 
 	c.tcpStateTable = &k8s.State{Table: make(map[int]*k8s.ServiceWithPort)}
 
-	c.deployLog = NewDeployLog()
```
jlevesy: Isn't `1000` a bit much for only debugging purposes ?

internal/controller/log.go
```diff
@@ -48,15 +51,25 @@ func (d *DeployLog) LogDeploy(timeStamp time.Time, podName string, podIP string,
 		Reason:           reason,
 	}
 
-	d.Entries = append(d.Entries, newEntry)
+	if len(d.entries) >= d.maxEntries {
+		// Pull elements off the front of the slice to make sure that the newly appended record is under the max record value.
+		d.entries = d.entries[(len(d.entries)-d.maxEntries)+1:]
```
jlevesy: Hmmm, the only possible case here is `len(d.entries) == d.maxEntries` right ? 
So `len(d.entries) - d.maxEntries == 0`  at any point ? 

Then this is correct ?
```suggestion
		d.entries = d.entries[1:]
```

internal/controller/log.go
jlevesy: Also where it is called except from tests ?

SantoDE: 👍

internal/controller/controller.go
SantoDE: Depends on the cluster size I guess?

internal/controller/log.go
dtomcej: I wanted to cover a theoretical edge case where somehow 2 entries got added before the slice was pruned, so it could have +2 entries etc.

internal/controller/controller.go
dtomcej: I figured this could be configurable if someone wanted to contribute and fiddle with the templates in the future.

internal/controller/log.go
```diff
@@ -40,23 +43,28 @@ func (d *DeployLog) Init() error {
...
-	if err != nil {
-		log.Error("Could not marshal deploylog entries")
-	}
+func (d *DeployLog) GetLog() []Entry {
+	return d.entries
+}
 
-	return data
+// Len returns the number of records in the entries slice.
+func (d *DeployLog) Len() int {
```
jlevesy: Is this `Len` function used somewhere ?

dtomcej: Ah it was used for https://github.com/containous/maesh/pull/342/files#diff-6edd2e4ad97b4bf62fa0c917f59bc634R54 previously, but not anymore I suppose!

jlevesy: Then I would remove it :)

PR #344, opened 2019-11-06
[DRAFT] Create Unit tests for the Controller and API via Mocks

dtomcej:
This PR so far creates a client mock, and creates a new controller object.

It modifies some interfaces to allow the mock to be used.

We use the client interfaces almost exclusively except in the controller code so far.

Once we have an initialized controller object, we can use it for unit tests.


Ullaakut:
@jlevesy proposed an API refactor, which makes a lot of sense, but would invalidate those tests. WDYT of waiting for the refactor and adding a `no-merge` on this one in the meantime?

dtomcej:
Closing in favour of #445

PR #345, opened 2019-11-06
Add API documentation

### LINKED ISSUE #327: Document API Endpoints
#### Proposal

Since we are implementing an API, we should document what the endpoints are

dtomcej:
This PR:

- Adds API documentation for the newly added Controller API

The documentation can be viewed via the netlify preview 😄 

Fixes #327

### DIFF
--- a/docs/content/api.md
+++ b/docs/content/api.md
@@ -33,5 +33,3 @@ Otherwise, it will return a 500.
 This endpoint provides a json array containing details about configuration deployments made by the controller.
 This array is currently capped at 1000 entries to avoid memory issues.
 If this is not enough, please open a github issue and we will look into updating this to be configurable.
-
-
--- a/docs/content/api.md
+++ b/docs/content/api.md
@@ -1,0 +1,37 @@
+# API
+
+Maesh includes a built-in API that can be used for debugging purposes.
+This can be useful when Maesh is not working as intended.
+The API is accessed via the controller pod, and for security reasons is not exposed via service.
+The API can be accessed by making a `GET` request to `http://<control pod IP>:9000` combined with one of the following paths:
+
+## `/api/configuration/current`
+
+This endpoint provides raw json of the current configuration built by the controller.
+
+!!! Note
+    This may change each request, as it is a live data structure.
+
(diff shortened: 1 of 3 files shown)


docs/content/api.md
```diff
@@ -0,0 +1,35 @@
...
+
+## `/api/status/node/{node}/configuration`
+
+This endpoint provides raw json of the current configuration on the Maesh node with the pod name given in `{node}`.
+This endpoint provides a 404 response if the pod cannot be found, or other non-200 status codes on other errors.
+If errors are encountered, the error will be returned in the body, and logged on the controller.
+
+## `/api/status/readiness`
+
+This endpoint returns a 200 response if the controller successfully deployed a configuration to a Maesh node, and Maesh is ready for use.
```
jlevesy: WDYT ?

```suggestion
This endpoint returns a 200 response if the controller successfully deployed a configuration to all Maesh nodes, and Maesh is ready for use.
```

docs/content/api.md
```diff
@@ -0,0 +1,35 @@
...
+Otherwise, it will return a 500.
+
+## `/api/log/deploylog`
```
jlevesy: `/api/log/deploylog` feels a bit redundant to me.
I'm asking a feature here haha, but WDYT of `/api/log` or `/api/log/deployment` instead ?

docs/content/api.md
```diff
@@ -0,0 +1,35 @@
...
+
+!!! Note
+    This may change each request, as it is a live data structure.
+
+## `/api/status/nodes`
+
+This endpoint provides a json array containing some details about the readiness of the Maesh nodes visible by the controller
+This endpoint will still return a 200 if there are no visible nodes.
+
+## `/api/status/node/{node}/configuration`
```
jlevesy: Not convinced by this change, but I would be nice if we had a hint about using the pod name directly.

```suggestion
## `/api/status/node/{maesh-pod-name}/configuration`
```

docs/content/api.md
SantoDE: Agree on that

PR #346, opened 2019-11-07
Bundle SMI CRD's with helm install

### LINKED ISSUE #217: Create option to Register SMI CRDs
We should provide the option to Register the SMI CRDs when installing Maesh with SMI enabled.

Otherwise, it installs correctly, but users cannot create SMI objects to configure Maesh with.

SantoDE:
Fixes #217 

That PR adds the option to install the SMI CRD's together with the normal installation of maesh.

### DIFF
--- a/helm/chart/maesh/templates/smi/smi-spec.yaml
+++ b/helm/chart/maesh/templates/smi/smi-spec.yaml
@@ -4,8 +4,6 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: httproutegroups.specs.smi-spec.io
-#  annotations:
-#    "helm.sh/hook": crd-install
 spec:
   group: specs.smi-spec.io
   version: v1alpha1
--- a/helm/chart/maesh/templates/smi/smi-spec.yaml
+++ b/helm/chart/maesh/templates/smi/smi-spec.yaml
@@ -1,0 +1,35 @@
+{{- if .Values.smi.deploy }}
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: httproutegroups.specs.smi-spec.io
+#  annotations:
+#    "helm.sh/hook": crd-install
+spec:
+  group: specs.smi-spec.io
+  version: v1alpha1
+  scope: Namespaced
+  names:
+    kind: HTTPRouteGroup
+    shortNames:
+      - htr
+    plural: httproutegroups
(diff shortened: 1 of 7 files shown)


helm/chart/maesh/templates/smi/smi-spec.yaml
```diff
@@ -0,0 +1,35 @@
+{{- if .Values.smi.deploy }}
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: httproutegroups.specs.smi-spec.io
+#  annotations:
```
mmatur: Could be removed?

helm/chart/maesh/values.yaml
```diff
@@ -68,7 +68,9 @@ metrics:
   prometheus:
     enabled: true
 
-smi: false
+smi:
+  deploy: true
+  enable: false
```
dtomcej: Won't you need to change the controller args for this? since now `.Values.smi` is not a bool anymore?

helm/chart/maesh/templates/smi/smi-spec.yaml
SantoDE: Solved

helm/chart/maesh/values.yaml
SantoDE: Solved.

helm/chart/maesh/values.yaml
```diff
@@ -68,7 +68,9 @@ metrics:
   prometheus:
     enabled: true
 
-smi: false
+smi:
+  deploy: true
```
dtomcej: Did we want to deploy this by default? I think we should _not_ deploy by default.

jlevesy: If I get this right, this means that on a blank cluster (without any SMI CRDs previously deployed) I would need to set both flags `smi.enable: true` and `smi.deploy: true`  to have maesh running in SMI mode.
To me changing this default to false introduces a risk to end up with users installing maesh in SMI mode without having the CRDs installed :sweat_smile: . Not super good.
What is the risk of redeploying the CRDs in that case ? It should be a noop right ?

Also WDYT of renaming `deploy` to `deploycrds` to make this variable a bit more self explainatory ?

SantoDE: Isn't it the other way around? SMI was false, which meant maesh was not started in SMI mode. Now, smi.enable is still false so no behavior has been changed. However, I liked the rename suggestion and adapted that :)

Issue #253, opened 2019-09-25
Allow Maesh to Run in a Singular Namespace


cbrintnall:
#### Proposal

 

Have Maesh label its own services `.maesh` so it doesn't need a unique namespace. From there Maesh can then identify its own services via the label, and other services without.

On service creation Maesh should add the label `service=maesh`, or a unique ID that it tracks internally (`id=<uuid4>`).

 #### Background

 
The issue comes from multi-tenant Kubernetes clusters. Although it is _still_ possible to spin up another namespace by requesting one from the cluster admins, removing the need to have a unique namespace for the `.maesh` services would allow multi-tenant clusters to run Maesh in one namespace, similar to how Traefik can run in a singular namespace.

 #### Workarounds

 

None currently, although going down the recommended path (spinning up a unique namespace for Maesh) is also possible. Difficulty depends on the clusters themselves.

#### Extra

This may also require / allow (if not already possible) Maesh to target specific namespaces to watch. 

More context can be found here on Discourse: `https://community.containo.us/t/why-does-maesh-need-to-be-in-its-own-namespace/1850/3`

dtomcej:
We could extend the `ignoreWrapper` to include labels that can also be used to ignore events.

jlevesy:
I'll work on this one

jlevesy:
Hey @cbrintnall,

> Have Maesh label its own services .maesh

Quick question, could you clarify what you meant by `label its own services` please ?
Do you mean creating a service with a name in `.maesh` (for example `whoami.maesh`) ?
If so, it is sadly impossible as service names are required to comply with the RFC DNS-1035 which excludes the possibility of having the `.` character in a name. 

Creating the following service

```
---
kind: Service
apiVersion: v1
metadata:
  name: whoami.maesh
  namespace: playground
spec:
  selector:
    app: whoami
  ports:
    - name: web
      protocol: TCP
      port: 80
      targetPort: 80
```

You get this error

```
The Service "whoami.maesh" is invalid: metadata.name: Invalid value: "whoami.maesh": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
```

Is that what you meant or did I miss something ? (which is completely possible :sweat_smile:)

Anyway, thanks for your contribution !

cbrintnall:
[...]

PR #348, opened 2019-11-08
Run in single namespace

jlevesy:
## What does this PR do

This PR makes Maesh able to run in a single namespace.

Fixes: #253 

## How to test this

Install maesh in the same namespace of your applications, and enjoy your shiny service mesh  as usual :tada:

## How does it work

Now, all generated mesh services are labeled with `app=maesh` too.
Then the controller is now able to a list of apps, right now `maesh` and `jaeger` are hardcoded, and we might want to change that. But it makes maesh ignore the maesh services. 
For the initial `createMaeshService` I used  [fields selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#supported-operators) instead of filtering everything in memory.

I also made the filtering more time efficient using a map instead of iterating over the services, as we can have a lot of services this can be important. 

## Additional information

I kind of hacked my way to get it working, that is definitely not a final implementation.
Also It doesn't support CNI just yet. 

Draft for the moment as CI is super broken and it is not finished.

### DIFF
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -103,7 +103,7 @@ func (c *Controller) Init() error {
 	c.api = NewAPI(c.apiPort, &c.lastConfiguration, c.deployLog, c.clients, c.meshNamespace)
 
 	if c.smiEnabled {
-		c.provider = smi.New(c.clients, c.defaultMode, c.meshNamespace, c.tcpStateTable, c.ignored)
+		c.provider = smi.New(c.clients, c.defaultMode, c.tcpStateTable, c.ignored)
 		// Create new SharedInformerFactories, and register the event handler to informers.
 		c.smiAccessFactory = smiAccessExternalversions.NewSharedInformerFactoryWithOptions(c.clients.SmiAccessClient, k8s.ResyncPeriod)
 		c.smiAccessFactory.Access().V1alpha1().TrafficTargets().Informer().AddEventHandler(c.handler)
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -239,8 +239,6 @@ func (c *Controller) createMeshServices() error {
 		return fmt.Errorf("unable to get maesh services: %v", err)
 	}
 
-	// Use a map here to have lookups in log(N), so were O(n*log(n)) and not O(n^2) on this algorithm.
-	// N being the amount of services in the cluster, this can be important.
 	maeshSvcsByName := reduceServicesByName(maeshSvcs)
 
 	for _, service := range potentialSvcs {
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -311,8 +311,7 @@ func (c *Controller) createMeshService(service *corev1.Service) error {
 				Name:      meshServiceName,
 				Namespace: c.meshNamespace,
 				Labels: map[string]string{
-					"app":       "maesh",
-					"component": "maesh-svc",
+					"app": "maesh",
 				},
 			},
 			Spec: corev1.ServiceSpec{
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -226,29 +226,34 @@ func (c *Controller) startInformers(stopCh <-chan struct{}) {
 }
 
 func (c *Controller) createMeshServices() error {
-	potentialSvcs, err := c.clients.GetServicesWithSelectors(
-		metav1.NamespaceAll,
-		c.ignored.LabelSelector(),
-		c.ignored.FieldSelector(),
-	)
+	// Because createMeshServices is called after startInformers,
+	// then we already have the cache built, so we can use it.
+	svcs := c.kubernetesFactory.Core().V1().Services().Lister()
+
+	sel, err := c.ignored.LabelSelector()
+	if err != nil {
+		return fmt.Errorf("unable to build label selectors: %v", err)
+	}
+
+	potentialSvcs, err := svcs.List(sel)
 	if err != nil {
 		return fmt.Errorf("unable to get services: %v", err)
 	}
 
-	// Because createMeshServices is called after startInformers,
-	// then we already have the cache built, so we can use it.
-	maeshSvcs := c.kubernetesFactory.Core().V1().Services().Lister().Services(c.meshNamespace)
-
 	for _, service := range potentialSvcs {
+		if c.ignored.IsIgnored(service.ObjectMeta) {
+			continue
+		}
+
 		log.Debugf("Creating mesh for service: %v", service.Name)
 
 		meshServiceName := c.userServiceToMeshServiceName(service.Name, service.Namespace)
 
-		_, err := maeshSvcs.Get(meshServiceName)
+		_, err := svcs.Services(c.meshNamespace).Get(meshServiceName)
 		if err == nil {
 			continue
 		}
-		// We're expecting an IsNotFound error here, to only create the maesh service if it isn't found.
+		// We're expecting an IsNotFound error here, to only create the maesh service if it does not exist.
 		if err != nil && !errors.IsNotFound(err) {
 			return fmt.Errorf("unable to check if maesh service exists: %w", err)
 		}
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -232,12 +232,12 @@ func (c *Controller) createMeshServices() error {
 
 	sel, err := c.ignored.LabelSelector()
 	if err != nil {
-		return fmt.Errorf("unable to build label selectors: %v", err)
+		return fmt.Errorf("unable to build label selectors: %w", err)
 	}
 
 	potentialSvcs, err := svcs.List(sel)
 	if err != nil {
-		return fmt.Errorf("unable to get services: %v", err)
+		return fmt.Errorf("unable to get services: %w", err)
 	}
 
 	for _, service := range potentialSvcs {
--- a/internal/controller/controller.go
+++ b/internal/controller/controller.go
@@ -23,6 +23,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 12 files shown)


internal/providers/kubernetes/provider.go
```diff
@@ -47,15 +45,15 @@ func New(client k8s.CoreV1Client, defaultMode string, meshNamespace string, tcpS
 func (p *Provider) buildRouter(name, namespace, ip string, port int, serviceName string, addMiddlewares bool) *dynamic.Router {
 	if addMiddlewares {
 		return &dynamic.Router{
-			Rule:        fmt.Sprintf("Host(`%s.%s.%s`) || Host(`%s`)", name, namespace, p.meshNamespace, ip),
+			Rule:        fmt.Sprintf("Host(`%s.%s.maesh`) || Host(`%s`)", name, namespace, ip),
```
SantoDE: Does that work with the hardcoded .maesh here? I'm unsure, I can be totaly wrong tough!

internal/providers/kubernetes/provider.go
```diff
@@ -47,15 +45,15 @@ func New(client k8s.CoreV1Client, defaultMode string, meshNamespace string, tcpS
...
 			EntryPoints: []string{fmt.Sprintf("http-%d", port)},
 			Middlewares: []string{serviceName},
 			Service:     serviceName,
 		}
 	}
 
 	return &dynamic.Router{
-		Rule:        fmt.Sprintf("Host(`%s.%s.%s`) || Host(`%s`)", name, namespace, p.meshNamespace, ip),
+		Rule:        fmt.Sprintf("Host(`%s.%s.maesh`) || Host(`%s`)", name, namespace, ip),
```
SantoDE: same

internal/providers/kubernetes/provider.go
jlevesy: Hmm, from what I saw, `.maesh` is already hardcoded [here](https://github.com/containous/maesh/blob/master/internal/k8s/client.go#L365).

internal/controller/controller.go
```diff
@@ -223,36 +223,52 @@ func (c *Controller) startInformers(stopCh <-chan struct{}) {
...
 		if err := c.createMeshService(service); err != nil {
-			return fmt.Errorf("unable to get create mesh service: %v", err)
+			return fmt.Errorf("unable to create mesh service: %v", err)
 		}
 	}
 
 	return nil
 }
 
+func reduceServicesByName(svcs []*corev1.Service) map[string]*corev1.Service {
```
dtomcej: Could we rename this `createServiceNameMap()`?

internal/providers/kubernetes/provider.go
dtomcej: This is the TLD. It is hardcoded for now. 😄

internal/controller/handler.go
```diff
@@ -48,7 +48,7 @@ func (h *Handler) OnAdd(obj interface{}) {
 	// assert the type to an object to pull out relevant data
 	switch obj := obj.(type) {
 	case *corev1.Service:
-		if h.ignored.IsIgnoredService(obj.Name, obj.Namespace) {
+		if h.ignored.IsIgnoredService(obj.Name, obj.Namespace, obj.GetLabels()["app"]) {
```
dtomcej: what happens if there is no app key?

dtomcej:
I ❤️ Where this is going.

Design LGTM.

Switching to review.

internal/controller/controller.go
```diff
@@ -223,36 +223,52 @@ func (c *Controller) startInformers(stopCh <-chan struct{}) {
...
+	)
 	if err != nil {
 		return fmt.Errorf("unable to get services: %v", err)
 	}
 
-	for _, service := range services {
-		if c.ignored.IsIgnoredService(service.Name, service.Namespace) {
-			continue
-		}
+	maeshSvcs, err := c.clients.GetServicesWithSelectors(c.meshNamespace, "app==maesh", "")
```
dtomcej: Could we codify this with

`c.getMeshServiceSelector()` or something like that instead of the "app==maesh" label? I'm thinking something that we could re-use in other packages if needed

internal/controller/controller.go
```diff
@@ -223,36 +223,52 @@ func (c *Controller) startInformers(stopCh <-chan struct{}) {
...
+
 		meshServiceName := c.userServiceToMeshServiceName(service.Name, service.Namespace)
 
-		for _, subservice := range services {
-			// If there is already a mesh service created, don't bother recreating
-			if subservice.Name == meshServiceName && subservice.Namespace == c.meshNamespace {
-				continue
-			}
+		if _, found := maeshSvcsByName[meshServiceName]; found {
+			log.Debugf("Skipping already created maesh service: %s", service.Name)
```
dtomcej: Not sure if this log is needed. It can really really spam the log in larger clusters.

internal/controller/controller.go
dtomcej: Could we rename this `makeServiceNameMap()` or `createServiceMapByName()` or something like that?

internal/controller/controller.go
```diff
@@ -300,6 +316,10 @@ func (c *Controller) createMeshService(service *corev1.Service) error {
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      meshServiceName,
 				Namespace: c.meshNamespace,
+				Labels: map[string]string{
+					"app":       "maesh",
+					"component": "maesh-svc",
```
dtomcej: Is this component label required if we already are only using the app label?

Also, could we not have these labels generated by a similar function referenced in:
https://github.com/containous/maesh/pull/348/files#diff-0d5beddb6a9b5c9dbe4fd5fe543b2471R235

So that a change to one would reflect in the other?

internal/controller/handler.go
dtomcej: Should we update this to just be:

`if h.ignored.IsIgnoredService(obj.GetObjectMeta())` and have the `IsIgnoredService` parse the meta for name/namespace and labels?

dtomcej: Then that way if we change how/what/why meta is parsed, we don't have to modify the handler each time?

internal/k8s/client.go
```diff
@@ -362,8 +362,8 @@ func (w *ClientWrapper) patchCoreDNSConfigMap(coreDeployment *appsv1.Deployment,
 maesh:53 {
     errors
     rewrite continue {
-        name regex ([a-zA-Z0-9-_]*)\.([a-zv0-9-_]*)\.maesh maesh-{1}-{2}.maesh.svc.%[1]s
-        answer name maesh-([a-zA-Z0-9-_]*)-([a-zA-Z0-9-_]*)\.maesh\.svc\.%[2]s {1}.{2}.maesh
+        name regex ([a-zA-Z0-9-_]*)\.([a-zv0-9-_]*)\.maesh %[3]s-{1}-{2}.%[3]s.svc.%[1]s
```
dtomcej: if we modify the maesh service name here, don't we also have to ensure that the Maesh Service is created with the same maesh name format?

https://github.com/containous/maesh/pull/348/files#diff-0d5beddb6a9b5c9dbe4fd5fe543b2471R273

internal/k8s/client.go
```diff
@@ -583,9 +585,16 @@ func (w *ClientWrapper) GetService(namespace, name string) (*corev1.Service, boo
 
 // GetServices retrieves the services from the specified namespace.
 func (w *ClientWrapper) GetServices(namespace string) ([]*corev1.Service, error) {
+	return w.GetServicesWithSelectors(namespace, "", "")
+}
+
+// GetServicesWithSelectors retrieves the services from the specified namespace and given selectors.
+func (w *ClientWrapper) GetServicesWithSelectors(namespace string, labelSelectors, fieldSelector string) ([]*corev1.Service, error) {
```
dtomcej: Does this need to be added to the interface/mock?

internal/k8s/ignore.go
```diff
@@ -31,8 +28,40 @@ func (i *IgnoreWrapper) AddIgnoredService(serviceName, serviceNamespace string)
...
+func (i *IgnoreWrapper) FieldSelector() string {
+	var selectors []string
+
+	for _, n := range i.Namespaces {
+		selectors = append(selectors, "metadata.namespace!="+n)
+	}
+
+	// We are loosing the filter by specifng namespace and service name here, but not sure if it's needed.
+	for _, n := range i.Services {
+		selectors = append(selectors, "metadata.name!="+n.Name)
```
dtomcej: Not sure this is correct.

Services have a name/namespace combination, so we need to have some more logic in building this selector.

For example:
```
Ignore:
Namespace: kube-system
Service: default/kubernetes
```

Would result in:
`metadata.namespace!=kube-system, metadata.name!=kubernetes`
Which is close, but not right. 

I am not sure if field selectors are right for those complex filters

internal/providers/kubernetes/provider.go
dtomcej: This is the TLD. It is hardcoded for now.  😄 we effectively hardcoded it earlier by mandating that the `maeshNamespace` == "maesh".

Now it is actually hardcoded 🔨

internal/controller/controller.go
jlevesy: There is a problem here, what if two services have the same name across two different namespaces ?

dtomcej: what if the map key is `<name>/<namespace>`?

dtomcej: or `map["name"]["namespace"]*corev1.Service`

dtomcej: then we can reference the service directly

jlevesy: Or I can use the service lister here, and not really have to build another index, WDYT ?

internal/controller/handler.go
jlevesy: Yes, I agree with that :)

internal/controller/controller.go
dtomcej: Yes. Do it!

internal/k8s/client.go
jlevesy: I'm  pretty sure I'll remove this at some point.

internal/controller/controller.go
jlevesy: > There is a problem here, what if two services have the same name across two different namespaces ?

Actually I'm working with services only from the maesh namespace service to it's fine to do this.

internal/k8s/ignore.go
jlevesy: Yeah, that's what I'm saying in the comment haha. I'll find a better way.

internal/controller/controller.go
jlevesy: > Is this component label required if we already are only using the app label?

Nope,

> Also, could we not have these labels generated by a similar function referenced...

Will see what I can do.

internal/k8s/client.go
jlevesy: Currently no tests relies on this call.  I would add this only if needed. WDYT ?

internal/controller/controller.go
jlevesy: > Also, could we not have these labels generated by a similar function referenced...

I actually removed the need of having a label for "app: maesh" label so no need to factorize this here, right ?

internal/k8s/client.go
jlevesy: How would you do this ?  Writing a test ?

internal/k8s/ignore.go
jlevesy: Ended up removing field selectors and only relying on cache to build the initial validation.

internal/k8s/client.go
jlevesy: I ended up removing it.

internal/k8s/client.go
dtomcej: Not sure. Maybe have a struct that holds the service name, and has 2 recievers, one that generates a mesh service name from the user service name, and the other that generates this regex?

That way we could test the struct to ensure that the names it generates match the regex it spits out?

internal/controller/controller.go
```diff
@@ -223,30 +224,42 @@ func (c *Controller) startInformers(stopCh <-chan struct{}) {
 }
 
 func (c *Controller) createMeshServices() error {
-	services, err := c.clients.GetServices(metav1.NamespaceAll)
+	// Because createMeshServices is called after startInformers,
+	// then we already have the cache built, so we can use it.
+	svcs := c.kubernetesFactory.Core().V1().Services().Lister()
```
dtomcej: Could we create a listWrapper struct that contains a map of listers, that we initialize in the controller `Init()`?

That way we could create an interface for the wrapper, and pass the interface to providers etc, and not have to use the kubeClient wrapper, therefore ensuring that we use the cache moving forward.

dtomcej:
This is looking _really_ good. I look forwards to working on a follow up to expand this!

internal/k8s/client.go
dtomcej: however we might want to have that in a separate PR (and out of the scope of this issue)

internal/controller/controller.go
dtomcej: Again, we might want to have that as a separate issue that we track and test separately in a new PR to prevent creep.

jlevesy: I'm not sure I want to do this in this PR, this means lots of work.

internal/providers/kubernetes/provider.go
Ullaakut: Wait I'm a bit late to the party but isn't hardcoding `maeshNamespace` to `maesh` kind of the opposite of the goal of this PR? 🤔

jlevesy: Not really. The hardcoded `.maesh`  is already something enforced by the DNS rewrites rules we set up. 

The goal of this PR is to make maesh ignores it's resources in the namespace it's running on.

PR #359, opened 2019-11-15
Fix panics in SMI & Kubernetes provider

### LINKED ISSUE #358: Panic when getting applicable traffic targets on nil endpoint list
### Bug

When given an invalid configuration, which has no endpoints, in `func (p *Provider) BuildConfig() (*dynamic.Configuration, error)`, at line 92, we do the following:

```go
applicableTrafficTargets := p.getApplicableTrafficTargets(base.GetEndpointsFromList(service.Name, service.Namespace, endpoints), trafficTargetsInNamespace)
```

Which runs `base.GetEndpointsFromList` and passes those endpoints to `p.getApplicableTrafficTargets`.

`GetEndpointsFromList` can return a nil pointer:

```go
// GetEndpointsFromList returns an endpoint from a list.
func GetEndpointsFromList(name, namespace string, endpointList []*corev1.Endpoints) *corev1.Endpoints {
	for _, endpoints := range endpointList {
		if endpoints.Name == name && endpoints.Namespace == namespace {
			return endpoints
		}
	}

	return nil
}
```

Which is then misused in `getApplicableTrafficTargets`:

```go
func (p *Provider) getApplicableTrafficTargets(endpoints *corev1.Endpoints, trafficTargets []*accessv1alpha1.TrafficTarget) []*accessv1alpha1.TrafficTarget {
	var result []*accessv1alpha1.TrafficTarget

	if len(endpoints.Subsets) == 0 { // Panic here.
		log.Debugf("No applicable TrafficTargets for service %s/%s: No endpoint subsets", endpoints.Namespace, endpoints.Name)
	}
```

And triggers the following panic error:

```go
...
github.com/containous/maesh/internal/providers/smi.(*Provider).getApplicableTrafficTargets(0xc0000f7ee8, 0x0, 0xc000286028, 0x1, 0x1, 0x1, 0x1, 0x0)
	/Users/ullaakut/Work/go/src/github.com/containous/maesh/internal/providers/smi/provider.go:201 +0x3a
github.com/containous/maesh/internal/providers/smi.(*Provider).BuildConfig(0xc0000f7ee8, 0x1, 0x1, 0xc0002dc000)
...
```

### How to reproduce this bug

Call `BuildConfig` with this configuration:

<details>
<summary>config.yaml</summary>

```yaml
kind: TrafficTarget
metadata:
  name: example-traffic-target
destination:
  kind: ServiceAccount
  name: api-service
  namespace: foo
sources:
  - kind: ServiceAccount
    name: prometheus
    namespace: default
specs:
  - kind: TCPRoute
    name: api-service-routes
---
kind: TCPRoute
metadata:
  name: api-service-routes

---
apiVersion: v1
kind: Service
metadata:
  name: test
  namespace: foo
  annotations:
    maesh.containo.us/traffic-type: tcp
spec:
  clusterIP: 10.1.0.1
  selector:
    app: test
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
---
apiVersion: v1
kind: Endpoints
metadata:
  name: demo-service
subsets:
- addresses:
  - ip: 10.1.1.50
    targetRef:
      name: example
      namespace: default
  ports:
  - port: 50

---
apiVersion: v1
kind: Pod
metadata:
  name: example
spec:
  serviceAccountName: api-service
  containers:
    - name: example
      image: busybox
status:
  podIP: "10.4.3.2"

---
apiVersion: v1
kind: Service
metadata:
  name: demo-service
spec:
  clusterIP: 10.1.0.1
  ports:
  - protocol: TCP
    port: 80
    name: web
```

</details>

Ullaakut:
### Goal of this PR

This PR fixes #358 by doing the following changes:

In `smi/provider.go`:

* Verify that endpoints are non-nil before accessing their subsets
* Verify that address target references are non-nil before accessing their name/namespace
* Verify that tcpStateTables are non-nil before accessing their table

In `kubernetes/provider.go`

* Verify that tcpStateTables are non-nil before accessing their table

### How to test it

Use the configuration files provided in the issue #358 in the provider unit tests and ensure that they pass. It should not be the case with the current state of master.

### Additional Notes

This PR will most likely be blocking for the next PR I plan on making, which introduces a unit test of building a TCP configuration in the SMI provider. It's while building the configuration that I found the issues described here, so it might just be that I misconfigured it, in which case the PR won't be blocking.

### DIFF
--- a/internal/providers/kubernetes/provider.go
+++ b/internal/providers/kubernetes/provider.go
@@ -241,6 +241,10 @@ func buildRateLimitMiddleware(annotations map[string]string) *dynamic.RateLimit 
 }
 
 func (p *Provider) getMeshPort(serviceName, serviceNamespace string, servicePort int32) int {
+	if p.tcpStateTable == nil {
+		return 0
+	}
+
 	for port, v := range p.tcpStateTable.Table {
 		if v.Name == serviceName && v.Namespace == serviceNamespace && v.Port == servicePort {
 			return port
--- a/internal/providers/smi/provider.go
+++ b/internal/providers/smi/provider.go
@@ -198,6 +198,11 @@ func (p *Provider) getTrafficSplitsWithDestinationInNamespace(namespace string, 
 func (p *Provider) getApplicableTrafficTargets(endpoints *corev1.Endpoints, trafficTargets []*accessv1alpha1.TrafficTarget) []*accessv1alpha1.TrafficTarget {
 	var result []*accessv1alpha1.TrafficTarget
 
+	if endpoints == nil {
+		log.Debugf("No applicable TrafficTargets: no endpoint")
+		return nil
+	}
+
 	if len(endpoints.Subsets) == 0 {
 		log.Debugf("No applicable TrafficTargets for service %s/%s: No endpoint subsets", endpoints.Namespace, endpoints.Name)
 	}
@@ -228,6 +233,11 @@ func (p *Provider) getApplicableTrafficTargets(endpoints *corev1.Endpoints, traf
 			var validPodFound bool
 
 			for _, address := range subset.Addresses {
+				if address.TargetRef == nil {
+					log.Error("Address has no target reference")
+					continue
+				}
+
 				pod, exists, err := p.client.GetPod(address.TargetRef.Namespace, address.TargetRef.Name)
 				if err != nil {
 					log.Errorf("Could not get pod %s/%s: %v", address.TargetRef.Namespace, address.TargetRef.Name, err)
@@ -536,6 +546,10 @@ func (p *Provider) buildTrafficSplit(config *dynamic.Configuration, trafficSplit
 }
 
 func (p *Provider) getMeshPort(serviceName, serviceNamespace string, servicePort int32) int {
+	if p.tcpStateTable == nil {
+		return 0
+	}
+
 	for port, v := range p.tcpStateTable.Table {
 		if v.Name == serviceName && v.Namespace == serviceNamespace && v.Port == servicePort {
 			return port
--- a/internal/providers/smi/provider.go
+++ b/internal/providers/smi/provider.go
@@ -230,32 +230,7 @@ func (p *Provider) getApplicableTrafficTargets(endpoints *corev1.Endpoints, traf
 				continue
 			}
 
-			var validPodFound bool
-
-			for _, address := range subset.Addresses {
-				if address.TargetRef == nil {
-					log.Error("Address has no target reference")
-					continue
-				}
-
-				pod, exists, err := p.client.GetPod(address.TargetRef.Namespace, address.TargetRef.Name)
-				if err != nil {
-					log.Errorf("Could not get pod %s/%s: %v", address.TargetRef.Namespace, address.TargetRef.Name, err)
-					continue
-				}
-
-				if !exists {
-					log.Errorf("pod %s/%s do not exist", address.TargetRef.Namespace, address.TargetRef.Name)
-					continue
-				}
-
-				if pod.Spec.ServiceAccountName == trafficTarget.Destination.Name {
-					validPodFound = true
-					break
-				}
-			}
-
-			if !validPodFound {
+			if !p.validPodFound(subset.Addresses, trafficTarget.Destination.Name) {
 				// No valid pods with serviceAccount found on the subset, so it is not affected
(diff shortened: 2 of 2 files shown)


Ullaakut:
Suggestions on how to break down this func:

```go
func (p *Provider) getApplicableTrafficTargets(endpoints *corev1.Endpoints, trafficTargets []*accessv1alpha1.TrafficTarget) []*accessv1alpha1.TrafficTarget {
	var result []*accessv1alpha1.TrafficTarget

	if endpoints == nil {
		log.Debugf("No applicable TrafficTargets: no endpoint")
		return nil
	}

	if len(endpoints.Subsets) == 0 {
		log.Debugf("No applicable TrafficTargets for service %s/%s: No endpoint subsets", endpoints.Namespace, endpoints.Name)
	}

	for _, subset := range endpoints.Subsets {
		for _, trafficTarget := range trafficTargets {
			if endpoints.Namespace != trafficTarget.Destination.Namespace {
				// Destination not in service namespace, skip.
				log.Debugf("Destination namespace for TrafficTarget: %s not in service namespace: %s", trafficTarget.Destination.Name, endpoints.Namespace)
				continue
			}

			var subsetMatch bool

			for _, endpointPort := range subset.Ports {
				if strconv.FormatInt(int64(endpointPort.Port), 10) == trafficTarget.Destination.Port || trafficTarget.Destination.Port == "" {
					subsetMatch = true
					break
				}
			}

			if !subsetMatch {
				// No subset port match on destination port, so subset is not affected
				log.Debugf("TrafficTarget: %s does not match destination ports for endpoints %s/%s", trafficTarget.Destination.Name, endpoints.Namespace, endpoints.Name)
				continue
			}

			var validPodFound bool

			for _, address := range subset.Addresses {
				if address.TargetRef == nil {
					log.Error("Address has no target reference")
					continue
				}

				pod, exists, err := p.client.GetPod(address.TargetRef.Namespace, address.TargetRef.Name)
				if err != nil {
					log.Errorf("Could not get pod %s/%s: %v", address.TargetRef.Namespace, address.TargetRef.Name, err)
					continue
				}

				if !exists {
					log.Errorf("pod %s/%s do not exist", address.TargetRef.Namespace, address.TargetRef.Name)
					continue
				}

				if pod.Spec.ServiceAccountName == trafficTarget.Destination.Name {
					validPodFound = true
					break
				}
			}

			if !validPodFound {
				// No valid pods with serviceAccount found on the subset, so it is not affected
				log.Debugf("Endpoints %s/%s has no valid pods with destination service account: %s", endpoints.Namespace, endpoints.Name, trafficTarget.Destination.Name)
				continue
			}

			// We have a subset match, and valid referenced pods for the trafficTarget.
			result = append(result, trafficTarget)
		}
	}

	return result
}
```

Without making it _less_ readable than with the 16 cyclomatic complexity it currently has?

jlevesy:
> Verify that tcpStateTables are non-nil before accessing their table.

Hmm, is that something possible ? TCP state table is always created [here](https://github.com/containous/maesh/blob/master/internal/controller/controller.go#L100) and passed to the providers 
[here](https://github.com/containous/maesh/blob/master/internal/controller/controller.go#L121). 

IMO no need to check for non-nil, this can't really happen. And if it happens we need to know it because it means something went really really wrong haha :sweat_smile:

dtomcej:
@Ullaakut could we pull this out to a method:

```
			var validPodFound bool

			for _, address := range subset.Addresses {
				if address.TargetRef == nil {
					log.Error("Address has no target reference")
					continue
				}

				pod, exists, err := p.client.GetPod(address.TargetRef.Namespace, address.TargetRef.Name)
				if err != nil {
					log.Errorf("Could not get pod %s/%s: %v", address.TargetRef.Namespace, address.TargetRef.Name, err)
					continue
				}

				if !exists {
					log.Errorf("pod %s/%s do not exist", address.TargetRef.Namespace, address.TargetRef.Name)
					continue
				}

				if pod.Spec.ServiceAccountName == trafficTarget.Destination.Name {
					validPodFound = true
					break
				}
			}
```
ad pass in `trafficTarget.Destination.Name` and `subset.Addresses`, and then have the method return if a pod that matches can be found?

Ullaakut:
@jlevesy I got the issue when experimenting with configs, so it seems like it's possible somehow 🤔 See https://github.com/containous/maesh/issues/358 

Might be something that can only happen with the mock client though.

PR #361, opened 2019-11-18
Experiment with multiple workers and workqueues.

jlevesy:
Hello hello,

### What does this PR do ? 

I spent some time thinking about this controller issue trying to split it.

To sum up what I've did:

I split the controller into two control loops, to make sure that one control loop does only one thing, this is what's recommended when writing a k8s controllers.

- The `mesher` loop, in charge of maintaining the collection of maesh services. 
- The `configurator` loop, in charge of watching every elements impacting the mesh nodes dynamic configuration and reconfigure all the mesh nodes. 

The second loop is highly subject to events (mostly because it actually depends/watches on a lot of resources) so I reintroduced back the workqueues in this controller which provides rate limiting easily in that case.
It also provides retries at event level, if the control loop cannot handle an error for some reason, then it gets requeued and will be processed after a while.

To finish, it fully relies on observers Listers.

WDYT of this ?

### How to test it ?

Build and update and image on your k8s cluster, and deploy a set of mesh nodes (applying this directory [here for instance](https://github.com/jlevesy/maesh-dev/tree/master/k8s/maesh)).

### Additional information

- It's a naive implementation, my goal is is demonstrate that this architecture works. It doesn't support TCP nor SMI.
- Test coverage is not complete.

### DIFF
--- a/internal/configurator/controller.go
+++ b/internal/configurator/controller.go
@@ -7,6 +7,7 @@ import (
 	"github.com/containous/maesh/internal/resource"
 	"github.com/containous/traefik/v2/pkg/config/dynamic"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/time/rate"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/informers"
 	listersv1 "k8s.io/client-go/listers/core/v1"
@@ -37,7 +38,11 @@ type Controller struct {
 
 // NewController returns a configurator controller.
 func NewController(informerFactory informers.SharedInformerFactory, provider base.Provider, deployer Deployer, currentNamespace string, l logrus.FieldLogger) *Controller {
-	q := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), moduleName)
+	q := workqueue.NewNamedRateLimitingQueue(
+		// TODO tweak this rate limiting
+		&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(10), 2)},
+		moduleName,
+	)
 
 	// Watch the mesh-services
 	svcInformer := informerFactory.Core().V1().Services()
@@ -50,9 +55,10 @@ func NewController(informerFactory informers.SharedInformerFactory, provider bas
 	// Watch all the endpoints.
 	endpointsInformer := informerFactory.Core().V1().Endpoints()
 	endpointsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    func(_ interface{}) { q.Add(struct{}{}) },
-		UpdateFunc: func(_, _ interface{}) { q.Add(struct{}{}) },
-		DeleteFunc: func(_ interface{}) { q.Add(struct{}{}) },
+		AddFunc: func(_ interface{}) { q.AddRateLimited(struct{}{}) },
+		// We rate limit the update func on endpoints, which can generate a lot of noise, when deploying pods.
+		UpdateFunc: func(_, obj interface{}) { q.AddRateLimited(struct{}{}) },
+		DeleteFunc: func(_ interface{}) { q.AddRateLimited(struct{}{}) },
 	})
 
 	// Watch the mesh nodes pods.
@@ -155,7 +161,7 @@ func enqueueSvcEvent(q workqueue.RateLimitingInterface, obj interface{}, log log
 		return
 	}
 
-	q.Add(struct{}{})
+	q.AddRateLimited(struct{}{})
 }
 
 func enqueuePodEvent(q workqueue.RateLimitingInterface, obj interface{}, log logrus.FieldLogger) {
@@ -168,5 +174,5 @@ func enqueuePodEvent(q workqueue.RateLimitingInterface, obj interface{}, log log
 		return
 	}
 
-	q.Add(struct{}{})
+	q.AddRateLimited(struct{}{})
 }
--- a/internal/configurator/controller.go
+++ b/internal/configurator/controller.go
@@ -1,0 +1,172 @@
+package configurator
+
+import (
+	"context"
+
+	"github.com/containous/maesh/internal/providers/base"
+	"github.com/containous/maesh/internal/resource"
+	"github.com/containous/traefik/v2/pkg/config/dynamic"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
+	listersv1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+)
+
+const moduleName = "configurator"
+
+// Deployer deploys a confiugration to a set of pods.
+type Deployer interface {
+	Deploy(pods []*corev1.Pod, cfg *dynamic.Configuration) error
+}
+
+// Controller is in charge of watching meaningfull resources and triggering mesh node reconfiguration.
+type Controller struct {
+	provider base.Provider
+	deployer Deployer
+	pods     listersv1.PodLister
+
+	currentNamespace string
+
+	queue  workqueue.RateLimitingInterface
+	exited chan struct{}
+
+	logger logrus.FieldLogger
+}
+
+// NewController returns a configurator controller.
+func NewController(informerFactory informers.SharedInformerFactory, provider base.Provider, deployer Deployer, currentNamespace string, l logrus.FieldLogger) *Controller {
+	q := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), moduleName)
+
+	// Watch the mesh-services
+	svcInformer := informerFactory.Core().V1().Services()
+	svcInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { enqueueSvcEvent(q, obj, l) },
+		UpdateFunc: func(_, obj interface{}) { enqueueSvcEvent(q, obj, l) },
+		DeleteFunc: func(obj interface{}) { enqueueSvcEvent(q, obj, l) },
+	})
+
+	// Watch all the endpoints.
+	endpointsInformer := informerFactory.Core().V1().Endpoints()
+	endpointsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(_ interface{}) { q.Add(struct{}{}) },
+		UpdateFunc: func(_, _ interface{}) { q.Add(struct{}{}) },
+		DeleteFunc: func(_ interface{}) { q.Add(struct{}{}) },
+	})
+
+	// Watch the mesh nodes pods.
+	podsInformer := informerFactory.Core().V1().Pods()
+	podsInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { enqueuePodEvent(q, obj, l) },
+		UpdateFunc: func(_, obj interface{}) { enqueuePodEvent(q, obj, l) },
+		DeleteFunc: func(obj interface{}) { enqueuePodEvent(q, obj, l) },
+	})
+
+	return &Controller{
+		deployer:         deployer,
+		provider:         provider,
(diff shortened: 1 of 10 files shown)


internal/configurator/controller.go
```diff
@@ -0,0 +1,178 @@
...
+	isMeshSvc, err := resource.IsMeshService(obj)
+	if err != nil {
+		log.WithError(err).Error("unable to determine if event comes from a mesh service, skipping")
+		return
+	}
+	if !isMeshSvc {
+		return
+	}
+
+	q.AddRateLimited(struct{}{})
```
Ullaakut: This is mysterious to me 🤔 Why do we add an empty struct to the rate limiting work queue? What does that do? Could we add a comment here to explain it?

internal/configurator/controller.go
```diff
@@ -0,0 +1,178 @@
...
+	isMeshPod, err := resource.IsMeshPod(obj)
+	if err != nil {
+		log.WithError(err).Error("unable to determine if event comes from a mesh pod, skipping")
+		return
+	}
+	if !isMeshPod {
+		return
+	}
+
+	q.AddRateLimited(struct{}{})
```
Ullaakut: Same here

internal/mesher/controller.go
```diff
@@ -0,0 +1,307 @@
...
+		svcCache:         svcInformer.Lister(),
+		svcClient:        clientSet.CoreV1().Services(ns),
+		queue:            q,
+		exited:           make(chan struct{}),
+		currentNamespace: ns,
+		logger:           l.WithField("module", moduleName),
+	}
+}
+
+// Run process all events coming from the workqueue until shutdown is signaled.
```
Ullaakut:
```suggestion
// Run processes events coming from the workqueue until shutdown.
```

internal/mesher/controller.go
```diff
@@ -0,0 +1,307 @@
...
+
+	queue  workqueue.RateLimitingInterface
+	exited chan struct{}
+
+	currentNamespace string
+
+	logger logrus.FieldLogger
+}
+
+// NewController setup event handlers on the service informer and returns a controller.
```
Ullaakut:
```suggestion
// NewController sets up event handlers on the service informer and returns a controller.
```

internal/mesher/controller.go
```diff
@@ -0,0 +1,307 @@
...
+
+// Run process all events coming from the workqueue until shutdown is signaled.
+func (c *Controller) Run() {
+	defer close(c.exited)
+
+	for c.processEvent() {
+	}
+}
+
+// ShutDown shutdowns the controller.
```
Ullaakut:
```suggestion
// ShutDown shuts the controller down.
```

internal/mesher/controller.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		c.handleError(item, err)
+		return true
+	}
+
+	c.queue.Forget(item)
+	return true
+}
+
+const (
```
Ullaakut:
```suggestion
const maxAttempts = 3
```

internal/mesher/controller.go
```diff
@@ -0,0 +1,307 @@
...
+		return fmt.Errorf("unable to lookup for maesh service: %w", err)
+	}
+
+	_, err = c.svcClient.Create(c.buildMeshSvc(meshSvcName, addedSvc))
+	if err != nil {
+		return fmt.Errorf("unable to create mesh service: %v", err)
+	}
+
+	c.logger.Infof(
+		"Created mesh service %q, because of the creation service %q in namespace %q",
```
Ullaakut: I'm guessing there's a mistake in this sentence 🤔

internal/mesher/controller.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		return fmt.Errorf("unable to retrieve old mesh svc from cache: %v", err)
+	}
+
+	newMeshSvc := c.buildMeshSvc(meshSvcName, updatedService)
+
+	newMeshSvc.ObjectMeta.ResourceVersion = oldMeshSvc.ObjectMeta.ResourceVersion
+	newMeshSvc.Spec.ClusterIP = oldMeshSvc.Spec.ClusterIP
+
+	_, err = c.svcClient.Update(newMeshSvc)
```
Ullaakut: Can be inlined

internal/mesher/controller.go
Ullaakut:
```suggestion
		"Created mesh service %q, because of the creation of service %q in namespace %q",
```

internal/mesher/controller_test.go
```diff
@@ -0,0 +1,217 @@
...
+			controller := mesher.NewController(clientSet, factory, meshNamespace, logger())
+
+			factory.Start(ctx.Done())
+
+			test.createResources(t, clientSet)
+			factory.WaitForCacheSync(ctx.Done())
+
+			go controller.Run()
+
+			// Forces the processing of the createResources events by the controller.
```
Ullaakut: This is waiting for processing more than forcing processing, right?

internal/resource/identifier.go
```diff
@@ -0,0 +1,79 @@
...
+// MeshPodsLabelsSelector returns the selector to apply when looking for maesh pods.
+func MeshPodsLabelsSelector() labels.Selector {
+	sel := labels.Everything()
+	sel = sel.Add(mustRequirement(labels.NewRequirement(AppLabel, selection.Equals, []string{AppLabelMaesh})))
+	sel = sel.Add(mustRequirement(labels.NewRequirement(ComponentLabel, selection.Equals, []string{ComponentLabelMeshNode})))
+	return sel
+}
+
+func mustRequirement(c *labels.Requirement, err error) labels.Requirement {
+	if err != nil {
```
Ullaakut: In which case could this happen?

dtomcej:
I have some general comments about this one:

1. The decoupling of the mesh service controller and the configuration controller can lead to situations where mesh services have been created, modified, and the configuration has to "catch up". This leads to a bad user experience.

2. The code will get a lot nastier when we implement SMI because they are all custom resources, and have their own clientsets, and will have to be managed separately.

I ask this question:

Does it make sense to create Maesh services before the proxy configuration is able to use them? If not, would you consider reducing back to one controller loop, where each event triggers a configuration rebuild?

For example: https://github.com/containous/maesh/pull/361/files#diff-b6032ec59f246b289de2ef416fa21f5aR114 Does not rely on `item` at all. `item` could be a boolean, or any value at all`

When looking at this, each item in the workqueue serves one purpose: To trigger a configuration rebuild, and a deploy of this configuration. Does it make sense to have a queue with a length of more than 1? What purpose does having a queue with 10 items serve?

Just some thoughts.

jlevesy:
> The decoupling of the mesh service controller and the configuration controller can lead to situations where mesh services have been created, modified, and the configuration has to "catch up". This leads to a bad user experience.

Could you elaborate in which case ? Maybe I missed something but, if the configuration loop is fast then it shouldn't really lag behind right ? Also, we have a periodic event (the resync from the informer) which makes sure Traefik nodes are reconfigured periodically in any case.

> The code will get a lot nastier when we implement SMI because they are all custom resources, and have their own clientsets, and will have to be managed separately.

It will get more complicated for sure, I was thinking of a `configurator.NewK8S(...)` and `configurator.NewSMI(...)` initialization functions for the controller to do the wiring, the we have to instantiate the right InformerFactories before that and pass the right listers into the SMI provider, but this shouldn't be super crazy complicated if we move the complexity into the initialization functions.

> Does it make sense to create Maesh services before the proxy configuration is able to use them? If not, would you consider reducing back to one controller loop, where each event triggers a configuration rebuild?

I agree it does not make sense per se. However, the two different controllers are actually watching two different (and exclusive) sets of resources and applies different reactions to those generated events.
- The `mesher` controller, watches all services that are not owned by the Maesh controller and makes sure that maesh services are in sync.
- The`configurator` controller, watches and each on each mutation it reconfigures the traefik nodes:
   - The maesh pods, if we add or update one ? (that's a lot, probably only add is fine), so we make sure all mesh pods are configured event if a new one pops. (exit the `deployToUnreadyPods`, and bonus it's way more reactive than a 10s poll).
   - The services owned by the maesh controller (create, update, delete), because the dynamic configuration has to be regenerated 
   - All the endpoints of the cluster (create, update and delete), because in that case the dynamic configuration needs also to be regenerated.

We also don't have the same rate limiting constraints on those queues, in the mesher case we want to handle every events. In the `configurator` not so much because it is exposed to a lot of events, so if we want to not spam the traefik nodes too much, then we might want to ratelimit those events.
The downside of that is that at the moment I have no strong proof that we are actually deploying the more up to date configuration, in the worst case it will be re-synced up because of the various observers resync period.

> For example: https://github.com/containous/maesh/pull/361/files#diff-b6032ec59f246b289de2ef416fa21f5aR114 Does not rely on item at all. item could be a boolean, or any value at all`

That's a naive implementation which is simple and does the job: on each event we recalculate the dynamic config and reconfigure each nodes.
We might want to be smarter that that if we need and be more specific per type of events (like dedicating a control loop to only configure newly added maesh nodes, specifically). 

> When looking at this, each item in the workqueue serves one purpose: To trigger a configuration rebuild, and a deploy of this configuration. Does it make sense to have a queue with a length of more than 1? What purpose does having a queue with 10 items serve?

The goal is to have in both cases a super quick control loop, and only rely on queue for their rate limiting capabilities, and also the retry of a failed event. As we will be relying on Informer cache in any case this should be fine for the processing part.

The unknown parameter are the external calls to the API server or the Traefik nodes. We should make super sure that they, in the worst case fail fast, so the event loop can go to the next event. And rely on the retry mechanism and the informer resync to make sure that all Traefik nodes are actually up to date.

WDYT?

internal/mesher/controller_test.go
jlevesy: To actually force the scheduler to execute the informers goroutines instead of this one. But I don't really like that and I want to improve it.

internal/resource/identifier.go
jlevesy: It's just an helper instead of handling error. This panic should never happen are those filters are right now not exposed to the user.

dtomcej:
@jlevesy when you install maesh in a cluster with 100 services and 10 pods per service (endpoints), The informer will receive the following events:

100 service creation events, 100 endpoint creation events, followed by 1000 endpoint update events.

So almost immediately there are going to be 1200 events in the queue to process. If a user creates a service right after installing maesh, The new events get put into the queue, and can be delayed while these initial events get "caught up".

Having stateless providers has relieved this issue by making sure that there is a full configuration loaded on the first event, and the next 1199 events result in a duplicate configuration and are skipped.

internal/configurator/controller.go
```diff
@@ -0,0 +1,178 @@
...
+		c.handleError(item, err)
+		return true
+	}
+
+	c.queue.Forget(item)
+
+	return true
+}
+
+const (
```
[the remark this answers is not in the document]
jlevesy: :woman_shrugging:, because probably way too much time awake in a plane :sweat:

internal/configurator/controller.go
jlevesy: This controller doesn't really care about what really happened, it just needs to recalculate dynamic configuration from current cache. So we don't provide any information about this event, only that it happened. Thus the empty struct.

internal/configurator/controller.go
jlevesy: https://github.com/containous/maesh/pull/361/files#r348810360

jlevesy:
> Having stateless providers has relieved this issue by making sure that there is a full configuration loaded on the first event, and the next 1199 events result in a duplicate configuration and are skipped.

Then my *naive* implementation has a flaw indeed :sweat_smile:, thanks.

In that case it will process those events, but it will recalculate the dynamic configuration based on the state of the informer cache, which is going to be immediately up to date on startup (because we called `WaitForCacheSync` on the informers factory). So we will send 1200 the same updated configuration to Traefik, (but at a limited pace thanks to rate limiting).

Which stills sucks, but nodes will have the correct configuration... However, we need to be smarter than that. 
On services and endpoints mutation compare the computed configuration with the last deployed to see if it makes sense or not to deploy.
When a node is added, then we send the send the configuration to this node no matter what.

WDYT ?

dtomcej:
Sounds good, that is what we currently do :)

https://github.com/containous/maesh/blob/master/internal/controller/controller.go#L166

PR #364, opened 2019-11-20
Upgrade helm to v3

### LINKED ISSUE #121: Re-evaluate helm

### LINKED ISSUE #293: Add support + examples for Helm v3
#### Proposal

Even with issue #121 on its way maybe consider add documentation and support for Helm v3. They have removed tiller so there are fewer security concerns involved.

 #### Background

I'm just getting started with Kubernetes, Helm, Maesh... there is so much to read, so much to learn. It's all a little bit overwhelming, tbh.

### LINKED ISSUE #296: Helm v2 being sunset
#### Proposal

Helm v2 will be sunset, upgrade to v3 chart

 #### Background

> As we move towards Helm 3.0.0, Helm 2 will be transitioning into "maintenance mode". Helm 2.15.0 is the final feature release for Helm 2. We will continue to accept bug fixes and fix any security issues that arise, but we will no longer be accepting new features for Helm 2. All feature development will be moved over to Helm 3.

https://github.com/helm/helm/releases/tag/v2.15.0

As helm v3 is still in beta, it is not critical to upgrade at this point.

### LINKED ISSUE #354: Remove Helm/Tiller from integration tests
#### Proposal

Since the removal of hooks from the chart, we should template out the yaml files, then apply them in the integration tests.

This means we can remove the tiller code, and not have to worry about waiting for helm to complete.

It will also make the helm v3 migration easier.

#### Background

`helm template .`

as part of the setup,

then in each suite we can just install Maesh via:

`kubectl apply -f .`

dtomcej:
This PR:

- Upgrades the version of helm to v3
- Removes tiller from integration test
- Removes templated CRDs
- Updates helm delete command
- *Actually* pin k3d install version

Notes:

This PR is incompatible with previous charts and versions of helm.

This PR will have to be merged timed with a release, as the documentation updates will also not be compatible with current live versions.

Fixes #296 
Fixes #293
Fixes #354
Fixes #121


README.md
```diff
@@ -42,14 +42,14 @@ To run this app, you require the following:
 
 - Kubernetes 1.11+
 - CoreDNS installed as [Cluster DNS Provider](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/) (versions 1.3+ supported)
-- Helm v2 with a [working tiller service account](https://helm.sh/docs/using_helm/#installing-tiller)
+- Helm v3
```
Ullaakut: So we won't be compatible with `v2` at all anymore?

helm/chart/maesh/values.yaml
```diff
@@ -69,7 +69,6 @@ metrics:
     enabled: true
 
 smi:
-  deploycrds: true
```
Ullaakut: Why this change?

integration/integration_test.go
```diff
@@ -238,35 +240,22 @@ func (s *BaseSuite) startWhoami(c *check.C) {
 	c.Assert(err, checker.IsNil)
 }
 
-func (s *BaseSuite) installTiller(c *check.C) {
-	// create tiller service account.
-	cmd := exec.Command("kubectl", "apply",
-		"-f", path.Join(s.dir, "resources/helm/serviceaccount.yaml"))
+func (s *BaseSuite) createMaeshNamespace(c *check.C) {
+	// Create maesh namespace, required by helm v3.
+	cmd := exec.Command("kubectl", "create", "namespace", maeshNamespace)
```
Ullaakut: Why not use the kube client library for the integration tests instead of `kubectl`? (I know it wasnt introduced in this PR but still, I'd like to know :p)

README.md
dtomcej: No, the helm v3 chart changes are not backwards compatible AFAIK

integration/integration_test.go
dtomcej: Very good question.

I will open a new issue to update and consolidate the commands to use the client lib

helm/chart/maesh/values.yaml
dtomcej: Because CRD deployment is handled within the helm command in v3, not via values.

integration/integration_test.go
dtomcej: #382

dtomcej:
>still having an issue with my Kops and kube-dns :(

Open an issue so we can fix it!

Issue #369, opened 2019-11-21
Implement Full Tracing Options


dtomcej:
#### Proposal

Maesh should provide full tracing options based on the Traefik Proxy.

This would require the following tracing providers:

 - [x] Jaeger
 - [x] Datadog
 - [ ] Zipkin
 - [ ] Instana
 - [ ] Haystack

PR #371, opened 2019-11-21
Add datadog tracing

dtomcej:
This PR:

- Adds datadog tracing to Maesh

Works towards #369

### DIFF
--- a/helm/chart/maesh/values.yaml
+++ b/helm/chart/maesh/values.yaml
@@ -59,8 +59,7 @@ tracing:
     enabled: true
     localagenthostport: ""
     samplingserverurl: ""
-  datadog:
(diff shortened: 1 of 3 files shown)


helm/chart/maesh/values.yaml
```diff
@@ -59,9 +59,18 @@ tracing:
     enabled: true
     localagenthostport: ""
     samplingserverurl: ""
+  datadog:
```
Ullaakut: Why not just leave it commented out by default instead of uncommented with an `enabled` value of `false`?

dtomcej: Sounds good. Will update #370 with the same.

PR #372, opened 2019-11-23
Make metrics subchart more configurable

### LINKED ISSUE #360: Metrics subchart does not use defaultStorageClass
https://github.com/containous/maesh/blob/85f4377551e7434155db62b2ce173c6a62fd261c/helm/chart/maesh/charts/metrics/values.yaml#L5

I tried installing Maesh with the default values in the Helm chart, but the PVCs never came alive due to the `local-path`. Does this require some additional setup or preparation beforehand? (I figure it may be the Rancher local-path-provisioner?)

valerauko:
This PR:
* fixes some typos in Guidlines.md
* makes it possible to pass StorageClassName to Grafana/Prometheus in the metrics subchart
* makes it possible to pass resource requests/limits to the metrics subchart (the default requests were pretty high, especially for Prometheus)

Fixes #360

### DIFF
--- a/helm/chart/maesh/values.yaml
+++ b/helm/chart/maesh/values.yaml
@@ -54,7 +54,7 @@ mesh:
 tracing:
   deploy: true
   jaeger:
-    image: 
+    image:
       name: groundnuty/k8s-wait-for:v1.2
     enabled: true
     localagenthostport: ""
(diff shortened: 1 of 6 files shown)


helm/chart/maesh/values.yaml
```diff
@@ -66,7 +66,20 @@ tracing:
 metrics:
   deploy: true
   prometheus:
+    # whether to expose Proometheus metrics
```
Ullaakut:
```suggestion
    # whether to expose Prometheus metrics
```

valerauko:
@dtomcej I see you're adding other metrics options -- is the way I did this PR fine with your changes?

dtomcej:
@valerauko it _should_ be fine.

The beauty of commenting out features by default is that it is easy to combine.

We may want to add the commented out values into the values.yaml as done in my PRs to illustrate what is possible instead of referring to other files, but its not a blocking requirement.

PR #379, opened 2019-12-02
Use initContainers to fix volume permissions

riker09:
When using Maesh with minikube the first time you run `helm install --name=maesh --namespace=maesh` everthing works nice. But minikube being a local development thingy it is quite common to `minikube stop` the cluster and then `minikube start` it again. Also known as rebooting. :wink: 

After the reboot the grafana and prometheus pods would not start and report some permission error. After digging around a while I figured the pods are run with a specific uid/gid and the volumes are mounted with root user id.

I'm not sure if I have provided a workaround or an actual fix. Maybe this PR needs more polishing. If so, please let me know.

### DIFF
--- a/helm/chart/maesh/charts/metrics/templates/grafana.yaml
+++ b/helm/chart/maesh/charts/metrics/templates/grafana.yaml
@@ -62,6 +62,16 @@ spec:
         - name: grafana-config
           mountPath: /etc/grafana
           readOnly: true
+      initContainers:
+      - name: metrics-storage-permission-fix
+        image: busybox
+        command: ["chown", "-R", "472:472", "/var/lib/grafana"]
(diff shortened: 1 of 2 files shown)


riker09:
I have tested the original chart without my changes against a k3s cluster. I can stop and start the k3s cluster without issues, so I believe this is an issue with minikube.

My k3s cluster is running with the `--docker` option. The default containerd has issues with pulling from GitHub Package Repository with imagePullSecrets.

SantoDE:
Hi @riker09 ,

thanks for doing that! I was actually about to do the same, so I can stop right here. Yeah, I tend to agree that this looks more like a workaround. 

I'll ping my colleagues and lets see what they feel about it :)

PR #389, opened 2019-12-13
Update Traefik to 2.1

### LINKED ISSUE #387: Update Traefik to 2.1
#### Proposal

We should update the images in Maesh and the api to use v2.1 since it is the new stable

#### Background

Update early and often to prevent debt.

SantoDE:
Fixes #387 

I bumped the integ tests as well as the Pod Image. Not really sure what you mean by api @dtomcej ? There have been no changes on the REST API for v2.1 iirc?

### DIFF
--- a/helm/chart/maesh/templates/_helpers.tpl
+++ b/helm/chart/maesh/templates/_helpers.tpl
@@ -18,7 +18,7 @@ Define the templated image with tag
 Define the templated mesh image with tag
 */}}
 {{- define "maesh.meshImage" -}}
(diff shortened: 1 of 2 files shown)


dtomcej:
I was just wanting to check to see if the v1alpha1 CRD version had been bumped as well.

Just because we need to ensure the `dynamic.config` version we create matches the version of the `dynamic.config` our Traefik pods expect :)

SantoDE:
Ah, I see! That's good tough @dtomcej All working as expected :)

Issue #385, opened 2019-12-10
Leverage Service Topology to prevent request node hops


dtomcej:
#### Proposal

Use service topology to prevent node hops for requests

#### Background

https://kubernetes-io-vnext-staging.netlify.com/docs/concepts/services-networking/service-topology/

This can be used to prevent node hops.

Related to #213

dtomcej:
This is currently waiting on k3s/k3d to have k8s version 1.17.

dtomcej:
https://github.com/rancher/k3s/releases/tag/v1.17.0%2Bk3s.1

dtomcej:
https://github.com/rancher/k3d/releases/tag/v1.5.1

PR #410, opened 2020-01-27
Update k3s to use k8s 1.17

dtomcej:
This PR:

- Updates k3d to 1.5.1, which uses k3s 1.17

This PR paves the way for dev on #385


dtomcej:
Blocking this on #400, since it doesn't make sense to merge this first.

PR #430, opened 2020-02-19
Enhance logging to show expects for TCPRoute and HTTPRoute

### LINKED ISSUE #417: TCPRoute and HTTPRoute in same namespace as TrafficTarget
#### Proposal

If we find that referenced TCPRoutes and HTTPRoutes are not in the same namespace, we should have a clearer error message rather than "Not found".

https://github.com/containous/maesh/blob/master/internal/providers/smi/provider.go#L361

#### Background

We don't log the expected values, we only log the error message returned. We should log the expected spec value (expected name/namespace) and the error.

SantoDE:
Fixes #417

### DIFF
--- a/internal/providers/smi/provider.go
+++ b/internal/providers/smi/provider.go
@@ -324,7 +324,7 @@ func (p *Provider) buildHTTPRouterFromTrafficTarget(serviceName, serviceNamespac
 
 		rawHTTPRouteGroup, err := p.httpRouteGroupLister.HTTPRouteGroups(trafficTarget.Namespace).Get(spec.Name)
 		if err != nil {
-			log.Errorf("Error getting the HTTPRouteGroups %s in the same Namespaceace %s as the TrafficTarget: %: %v", spec.Name, trafficTarget.Namespace, err)
+			log.Errorf("Error getting the HTTPRouteGroups %s in the same namespace %s as the TrafficTarget: %v", spec.Name, trafficTarget.Namespace, err)
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


internal/providers/smi/provider.go
```diff
@@ -360,7 +360,7 @@ func (p *Provider) buildTCPRouterFromTrafficTarget(trafficTarget *access.Traffic
 
 		_, err := p.tcpRouteLister.TCPRoutes(trafficTarget.Namespace).Get(spec.Name)
 		if err != nil {
-			log.Errorf("Error getting TCPRoute: %v", err)
+			log.Errorf("Error getting the TCPRoute %s in the same Namespaceace %s as the TrafficTarget: %v", spec.Name, trafficTarget.Namespace, err)
```
dtomcej:
```suggestion
			log.Errorf("Error getting the TCPRoute %s in the same namespace %s as the TrafficTarget: %v", spec.Name, trafficTarget.Namespace, err)
```

internal/providers/smi/provider.go
```diff
@@ -324,7 +324,7 @@ func (p *Provider) buildHTTPRouterFromTrafficTarget(serviceName, serviceNamespac
 
 		rawHTTPRouteGroup, err := p.httpRouteGroupLister.HTTPRouteGroups(trafficTarget.Namespace).Get(spec.Name)
 		if err != nil {
-			log.Errorf("Error getting HTTPRouteGroup: %v", err)
+			log.Errorf("Error getting the HTTPRouteGroups %s in the same Namespaceace %s as the TrafficTarget: %: %v", spec.Name, trafficTarget.Namespace, err)
```
dtomcej:
```suggestion
			log.Errorf("Error getting the HTTPRouteGroups %s in the same namespace %s as the TrafficTarget: %: %v", spec.Name, trafficTarget.Namespace, err)
```

PR #431, opened 2020-02-19
Add a note about being a namespaced ressource

### LINKED ISSUE #416: TrafficTarget Source and Destination namespace requirements
#### Proposal

We should document that the TrafficTarget is a namespaced resource, and that the source and destination namespaces need to be explicitly defined.

#### Background

Leaving the destination namespace empty does not work properly, because the destination namespace cannot be inferred from the TrafficTarget namespace, as empty _may_ be intended to mean the default namespace.

#### Workarounds

Explicitly require traffictarget namespace to be defined.

SantoDE:
Fixes #416

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -125,6 +125,9 @@ Other types of route groups and detailed information are available [in the speci
 
 By default, all traffic is denied so we need to grant access to clients to our application. This is done by defining a `TrafficTarget`.
 
+??? Note "TrafficTarget Source & Destination"
+    Please note, that TrafficTarget is a namespaced ressource. Therefore, the source and the destination namespace needs to be explicitly defined.
+
 ```yaml
 ---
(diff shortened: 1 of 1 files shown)


docs/content/configuration.md
```diff
@@ -125,6 +125,9 @@ Other types of route groups and detailed information are available [in the speci
 
 By default, all traffic is denied so we need to grant access to clients to our application. This is done by defining a `TrafficTarget`.
 
+??? Note "TrafficTarget Source & Destination"
+    Please note, that TrafficTarget is a namespaced ressource. Therefore, the source and the destination namespace needs to be explicitly defined.
```
dtomcej:
```suggestion
    Please note that TrafficTarget is a namespaced resource.
    Therefore, the source and the destination namespace needs to be explicitly defined.
```

PR #433, opened 2020-02-19
Increase legacy integration test timeouts

dtomcej:
This PR:

- Increases the legacy integration test timeouts to match the old CI

May reduce the fails on the v1.1 branch.


dtomcej:
This is not required, as we are not maintaining the old tests past v1.1

PR #439, opened 2020-02-24
Extract TCPStateTable logic from the controller

### LINKED ISSUE #427: Move TCP State Table outside of the Controller and Providers.
Blocked by: #425 

#### Proposal

We would like to introduce a new interface to abstract the reading and writing of the TCPStateTable in Maesh. 

BTW WDYT of renaming TCPStateTable into `TCPPortMapping` ?

#### Background

To use Maesh code in a different environment, we would like to use our own storage backend for the TCP State Table.

#### Workarounds

Unfortunately no.

jspdown:
This PR abstracts the logic behind the `controller.tcpStateTable` so Maesh code can be reused without the need to persists these data in a k8s ConfigMap. Providers now accepts a `TCPPortMapper` from which the can retrieve a port mapping for a given service. An implementation of this interface has been provided: TCPPortMapping.

Closes #427 

While extracting this component from the controller we have noticed that the controllers and providers where not checking if a port was in the range defined by `limits.tcp` and `limits.http` (from `values.yaml`). This issue has been addressed in this PR. 

Co-authored By: Landry Benguigui landry@containo.us


pkg/k8s/tcpportmapping.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+		return i, nil
+	}
+
+	return 0, errors.New("unable to find an available port")
+}
+
+func (m *TCPPortMapping) loadState() error {
+	cfg, err := m.client.CoreV1().ConfigMaps(m.cfgMapNamespace).Get(m.cfgMapName, metav1.GetOptions{})
```
dtomcej: Should use lister, not client getter.

pkg/k8s/tcpportmapping.go
```diff
@@ -0,0 +1,177 @@
...
+
+			m.table[int32(port)] = svc
+		}
+	}
+
+	return nil
+}
+
+func (m *TCPPortMapping) saveState() error {
+	cfg, err := m.client.CoreV1().ConfigMaps(m.cfgMapNamespace).Get(m.cfgMapName, metav1.GetOptions{})
```
dtomcej: Should use lister

pkg/providers/kubernetes/provider.go
```diff
@@ -17,13 +18,20 @@ import (
 	listers "k8s.io/client-go/listers/core/v1"
 )
 
+// TCPPortMapper is capable of retrieving a TCP port mapping for a given service.
+type TCPPortMapper interface {
```
dtomcej: Why is this interface different than for the controller?

pkg/providers/smi/provider.go
```diff
@@ -24,10 +25,15 @@ import (
 	listers "k8s.io/client-go/listers/core/v1"
 )
 
+// TCPPortMapper is capable of retrieving a TCP port mapping for a given service.
+type TCPPortMapper interface {
+	Find(svc k8s.ServiceWithPort) (int32, bool)
```
dtomcej: Are there any issues with having this interface less restrictive?

pkg/providers/smi/provider.go
```diff
@@ -77,6 +88,8 @@ func New(defaultMode string, tcpStateTable *k8s.State, ignored k8s.IgnoreWrapper
 
 // BuildConfig builds the configuration for routing
 // from a native kubernetes environment.
+// FIXME: Reduce cognitive complexity.
```
dtomcej: Is this FIXME required at this point?

This is our one method that is big :(

pkg/providers/smi/provider.go
```diff
@@ -162,6 +191,27 @@ func (p *Provider) BuildConfig() (*dynamic.Configuration, error) {
 	return config, nil
 }
 
+func (p *Provider) getTCPPort(namespace, name string, port int32) (int32, error) {
```
dtomcej: since this is required for multiple providers could it be moved to the base?

That way the kubernetes and SMI providers could use it.

.golangci.toml
```diff
@@ -39,6 +39,7 @@
     "scopelint",
     "funlen",
     "bodyclose", # Too many false positives
+    "godox",
```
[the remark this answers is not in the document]
jspdown: This can be debated but I personally don't see `FIXME`, `NOTE` or `TODO` as something that should trigger an error in the CI. For example here: https://github.com/containous/maesh/pull/439/files#diff-db68cf5cf70ac312d2dbd378e1e18b74R91

pkg/controller/controller.go
```diff
@@ -38,6 +38,13 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
+// TCPPortMapper is capable of storing and retrieving a TCP port mapping for a given service.
+type TCPPortMapper interface {
+	Find(svc k8s.ServiceWithPort) (int32, bool)
+	Get(srcPort int32) *k8s.ServiceWithPort
+	Add(svc *k8s.ServiceWithPort) (int32, error)
+}
```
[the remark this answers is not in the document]
jspdown: That's a good observation! We have noticed mappings were never removed nor updated when receiving a Update events or a Delete event from an informer. This is definitely something that should be done. WDYT about fixing this issue in a different PR (Since this issue was already present in the code before)?

pkg/providers/kubernetes/provider.go
jspdown: A provider shouldn't be able to Add a new mapping, it should only be capable of Finding an existing one. This is why we chose to expose only this method.

pkg/providers/smi/provider.go
jspdown: As answered [here](https://github.com/containous/maesh/pull/439/files#r383374288).

pkg/providers/smi/provider.go
jspdown: It's not required, but we felt really bad about adding the `// nolint:gocognit,gocyclo`. So we wanted to add this FIXME, to make sure we never ignore this `nolint` and make it clear this function should be refactored. WDYT?

pkg/controller/controller.go
```diff
@@ -38,6 +38,13 @@ import (
 	"k8s.io/client-go/util/retry"
 )
 
+// TCPPortMapper is capable of storing and retrieving a TCP port mapping for a given service.
+type TCPPortMapper interface {
```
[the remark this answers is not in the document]
jspdown: `MeshController` seems to be the only type which accepts this interface. Just curious, why could you put it in a separate file?

pkg/providers/smi/provider.go
dtomcej: 10-4

.golangci.toml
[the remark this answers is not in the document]
dtomcej: Okee.

pkg/controller/controller.go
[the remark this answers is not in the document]
dtomcej: Yup. Can you open an issue for that?

pkg/controller/controller.go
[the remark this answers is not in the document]
dtomcej: I see that you use different `TCPPortMapper` interfaces in different packages.

I'm wondering if it makes any sense to consolidate them.

pkg/providers/kubernetes/provider.go
dtomcej: Then should we use the same interface name as the interface that does have other methods?

pkg/controller/controller.go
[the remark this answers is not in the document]
jspdown: I see, `TCPPortMapper` defined in providers should maybe have a different name. We want to make sure a Provider can only Find a mapping and not add new one.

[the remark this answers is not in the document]
dtomcej: Finally figured out what you were doing with the interfaces.

I'm just an idiot sometimes.

pkg/providers/kubernetes/provider.go
jspdown: It definitely makes sense. I'm not convinced by `TCPPortMapperReadOnly`. What about `TCPPortFinder`? :)

dtomcej: Sounds good.

pkg/controller/controller.go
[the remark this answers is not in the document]
jspdown: An issue has been opened: #442

pkg/k8s/tcpportmapping.go
jspdown: Yes, we can use a lister. But if we do so, we shouldn't create the lister from the shared informer `kubernetesFactory` and instead create a new informer especially for this resource. If we reuse the same informer, like we did before, here's what is going to happen:
- `controller.StartInformer`: We start all informers and wait for there cache to be filled. At this point we start receiving events on the `handler`.
- `handler.OnAdd` gets called which calls `controller.createMeshService`. This function relies on the TCPStateTable, which is not initialized yet.
So, the TCPStateTable needs to be initialized before we receive events on the informer but after startInformer since we need to informer to be populated. This is not something we can do unless creating a different informer.

We chose to not use an informer because we were doing `Get`s and not `List`s, only when initializing the state and only when `Add` gets called, which is not very often (except on startup). Since We also call `Update` in the `Add`, it means that every time we call `Add` the informer will receive the updated configmap. So, we thought, not using an informer would result in (almost) the same amount of traffic and calls would be at the same frequency. One thing that motivated us to not use Informers in the case is that we are interested by a single configmap, so the informer index would be full of unwanted objects.

pkg/providers/smi/provider.go
jspdown: `getTCPPort` and `getHTTPPort` are not helpers, they both relies on the Provider' state.

pkg/controller/controller.go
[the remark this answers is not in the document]
jspdown: Ahah no worries, their names were not very intuitive :sweat_smile:

pkg/k8s/tcpportmapping.go
dtomcej: Makes sense.
We have been focusing on using listers for performance reasons, but you are right, we are only dealing with one configmap, so `Get()` should be fine.

PR #440, opened 2020-02-24
Add host to API bind

### LINKED ISSUE #404: Allow API to bind to IP, rather than all interfaces
Allowing the API to bind to all interfaces can cause interferance, and can trigger firewall issues on Macs.

It is probably a good idea to have an IP to bind to, so that integration tests can be run easier, and security can be improved in the future.

dtomcej:
This PR:

- Adds the ability to define a host/IP for the API to bind to, not just ports.

Fixes #404 

This prevents warnings and error popups on Mac integration tests (and some unit tests)

### DIFF
--- a/cmd/configuration.go
+++ b/cmd/configuration.go
@@ -14,6 +14,7 @@ type MaeshConfiguration struct {
 	Namespace        string   `description:"The namespace that maesh is installed in." export:"true"`
 	IgnoreNamespaces []string `description:"The namespace that maesh should be ignoring." export:"true"`
 	APIPort          int32    `description:"API port for the controller" export:"true"`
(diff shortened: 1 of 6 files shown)


cmd/configuration.go
```diff
@@ -14,6 +14,7 @@ type MaeshConfiguration struct {
 	Namespace        string   `description:"The namespace that maesh is installed in." export:"true"`
 	IgnoreNamespaces []string `description:"The namespace that maesh should be ignoring." export:"true"`
 	APIPort          int      `description:"API port for the controller" export:"true"`
+	APIIP            string   `description:"API IP for the controller to bind to" export:"true"`
```
jspdown: WDYT of renaming this to `APIHost` instead?

PR #461, opened 2020-03-09
Add CoreDNS Readiness check and Update Traefik Version

dtomcej:
This PR:

- Adds a CoreDNS readiness check
- Sets CoreDNS 1.6.3 as the default
- Updates the Traefik version used to match the version used in the controller


dtomcej:
This is bloody slow now too.

Need to speed it up.

Issue #464, opened 2020-03-11
Update logger to use FieldLogger interface


dtomcej:
#### Proposal

We should use the FieldLogger interface, since it is shared with a bunch of projects, and assists with integration of such.

#### Background

Trying to use maesh packages with other packages from other projects can cause logging issues.

PR #465, opened 2020-03-11
Use logrus.FieldLogger for logging

dtomcej:
This PR:

- Adds a FieldLogger interface to the following packages:
  - Prepare
  - k8s client
  - Controller

Further logging updates will be needed, but will be done in a follow up PR

Works towards fixing #464


ldez:
I'm not sure about this change.

dtomcej:
Issue has been updated to reflect proper goal.

PR #468, opened 2020-03-16
Fix port-mapping cleanup for HTTP services

jspdown:
This PR fixes the port-mapping cleanup method so it doesn't print a warning when cleaning up an HTTP service.

The service mode (http or tcp) wasn't checked. The TCP state table, being used only for TCP, when an HTTP service was cleaned up, a warning was displayed saying we couldn't remove a port mapping. This PR makes sure we are doing this only for TCP services. No cleanup is required on HTTP services.


dtomcej:
This appears to be stuck on broken netlify checks

PR #473, opened 2020-03-18
Allow users to limit what service ports are exposed

### LINKED ISSUE #330: Allow users to specify which ports are exposed via Maesh
#### Proposal

Currently Maesh exposes all ports, and does not allow users to specify which ports should be exposed or used. We should allow this to be configurable via the service.

dtomcej:
This PR:

- Checks annotations for matching ports, and only creates shadow services with those ports

TODO:

- Update providers to only create routers on matching ports

Fixes #330

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -26,6 +26,22 @@ Dynamic configuration can be provided to Maesh using either annotations on kuber
 
 ### With Kubernetes Services
 
+#### Exposed ports
+
+By default, all ports on services are exposed by maesh.
+If you would like to define which ports are exposed, you can use the following annotation:
+
+```yaml
+maesh.containo.us/exposed-ports: "http, admin, 8553"
+```
+
+The list of ports can be names or port numbers.
+Any ports that match will be exposed via maesh.
+
+??? Note "TargetPorts"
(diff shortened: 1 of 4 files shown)


dtomcej:
This is currently blocked on major provider updates. This first portion will be unaffected.

jspdown:
@dtomcej How different is TrafficTarget's `Destination.Port` from this new annotation?

dtomcej:
@jspdown, not much! except that this annotation should allow multiple ports to be opened, and to use named ports, where destination in traffictarget only allows one, and is defined by number only.

This also doesn't require the SMI CRDs, but it may be less of an issue if we are reworking the providers.

That is why I paused dev on this until I review the upcoming work :)

kevinpollet:
@dtomcej are you still planning to work on this as #490 has been merged?

dtomcej:
@kevinpollet @jspdown I'm wondering if this is of any value now.

Originally this would have been a bit of a stop-gap for non-SMI users, but I don't think that this is needed at this point in time.

What do you think?

kevinpollet:
Agree with you, I find this feature very interesting but I would like to see how things are going before adding it, e.g. after we work on the `traffic-type` annotation.

If @jspdown is ok, I'll close this PR.

PR #486, opened 2020-04-06
Remove old dependencies

dtomcej:
This PR:

- Removes rancher dependency that we don't need to fork
- Makes docker version consistent with Traefik
- Downgrades the SMI SDK to 0.3.0
- Downgrades the Client-Go and associated libs to 0.17.4
- Adds containous/mux fork
- Fixes go-http-auth version to match with Traefik


dtomcej:
Should we downgrade the `client-go` to 0.17.4 and move the SMI SDK to 0.3.0 for easier integration?

Ullaakut:
If there's something I can do to help with the downgrading, let me know :)

dtomcej:
Waiting on https://github.com/containous/mux/pull/7 to re-add the `containous/mux` fork

dtomcej:
Passing tests, ready for 👀

Issue #487, opened 2020-04-06
Uninstall Helm Chart should remove shadow services


jspdown:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

- Install Maesh using Helm
- Create a new Service (which should create a shadow-service)
- Uninstall the Helm chart

### What did you expect to see?

I was expected all resources created by Maesh to be removed. 

### What did you see instead?

Shadow services haven't been removed.

### Output of controller log: 

Not relevant for this bug report.

### What is your environment & configuration (arguments, provider, platform, ...)?

Maesh built from sources on master.

### If applicable, please paste the yaml objects required to reproduce your issue

Not relevant for this bug report.

jspdown:
[Charts Hooks](https://helm.sh/docs/topics/charts_hooks/) could be the solution.

dtomcej:
The chart can only remove services that are created by the chart. The shadow services are not created by the chart, and cannot be directly removed.

However,

We could have a `maesh cleanup` command that removes all the created objects, and is triggered from an uninstall hook!

jspdown:
@dtomcej Yes, exactly. That what I had in mind. Uninstalling Maesh and then re-installing it using Helm with the current version leads to undesired behavior on TCP services. Indeed, shadow services are created only if they don't exist, and so, the TCPPortMapper won't store the port mapping for the existing shadow service.

PR #490, opened 2020-04-07
ACL flag with reworked SMI/Service handling

### LINKED ISSUE #447: [Proposal] Rework SMI / Service Handling to have an ACL Flag
#### Proposal

I'd like to "merge" the behaviour of SMI / Service Configuration Discovery, as both are not exclusive to each other except the ACL (Traffic Target). Therefore, I'd like to change the current SMI Flag to an "ACL" flag in order to either disable or enable ACL.

#### Background

Currently, the user has to decide to either do SMI or normal mode. This leads to the fact, that currently, a user could either have ACL through TrafficTargets or circuit-breaking / retries. However, SMI mode should not be exclusive as this only standardized some part of service mesh configuration. Most of that parts (e.g. Traffic Splits) are not even "dangerous", as they're just a way to edits an app internal behavior. The only "dangerous" feature through SMI is ACL, as that could either deny or allow certain connections.

Therefore, switching to a more "feature orientated" switch rather than enable / disable the whole SMI set could make sense.

#### Workarounds

None so far.

jspdown:
### What does this PR do?

Closes #447 

This PR adds support for the ACL flag and rework the SMI/Service handling.

Currently, when the user installs Maesh he has to choose between 2 modes: Kubernetes service and SMI mode. Each of them provides different features:
- Kubernetes Service mode: Allow to use annotations on services (`circuit-breaker`, `rate-limiter` or `retry`).
- SMI mode: Allow to use SMI resources such as `TrafficSplit` to split the traffic coming to a service and `TrafficTarget` to authorize a set of pods to hit another set of pods.

This rewrite of the Provider will now allow us to use all these features at the same time, depending on if ACL mode is on or off:

 | Features              | ACL disabled | ACL enabled |
 |-----------------------|--------------|-------------|
 | Traffic-Type          | ✔            | ✔           |
 | Scheme                | ✔            | ✔           |
 | Retry                 | ✔            | ✔           |
 | Circuit-Breaker       | ✔            | ✔           |
 | Rate-Limit            | ✔            | ✔           |
 | Traffic-Split (SMI)   | ✔            | ✔           |
 | Traffic-Target (SMI)  | ✘            | ✔           |

As this is not intended to be a breaking change, the `smi` flag is still available. If the `acl` flag is not set but `smi` flag is set, it will behave like if the `acl` flag was set. 

### How to test it

- `make test`
- `make build`
- `docker tag containous/maesh:latest maesh-dev`

You can find some configuration files for different basic scenarios in the following archive:
[provider-test-assets.tar.gz](https://github.com/containous/maesh/files/4444460/provider-test-assets.tar.gz)

Test with ACL disabled:
- k3d create --server-arg --no-deploy --server-arg="traefik"`
- `k3d import-images maesh-dev`
- Install Maesh via Helm with the `acl=false` flag (or without it):
```
  --namespace=maesh \
  --set controller.image.pullPolicy=IfNotPresent \
  --set controller.image.name=maesh-dev \
  --set controller.image.tag=latest \
  --set limits.tcp=20 \
  --set limits.http=20 \
  --set smi.deploy=true \
  --set acl=false
```
- `kubectl apply -f provider-test-assets/acl-disabled/http`

Test wit ACL enabled:
- Make sure you start from a clean state (issue #487).
- Install  Maesh via Helm with the `acl=true` flag:
```
  --namespace=maesh \
  --set controller.image.pullPolicy=IfNotPresent \
  --set controller.image.name=maesh-dev \
  --set controller.image.tag=latest \
  --set limits.tcp=20 \
  --set limits.http=20 \
  --set smi.deploy=true \
  --set acl=true 
```
- `kubectl apply -f provider-test-assets/acl-enabled/http`

### Additional Notes

Co-authored-by: Landry Benguigui landry@containo.us


Ullaakut:
Will give this some 👀 today.

helm/chart/maesh/templates/controller/controller-deployment.yaml
```diff
@@ -70,6 +70,9 @@ spec:
             {{- if .Values.mesh }}
             - "--defaultMode={{ .Values.mesh.defaultMode }}"
             {{- end }}
+            {{- if .Values.acl }}
```
dtomcej, line 73:
```suggestion
            {{- if or .Values.acl .Values.smi.enable}}
```

dtomcej: Since we are deprecating the `--smi` flag

helm/chart/maesh/templates/controller/controller-deployment.yaml
```diff
@@ -70,6 +70,9 @@ spec:
...
+            - "--acl"
+            {{- end }}
             {{- if .Values.smi.enable }}
```
dtomcej, line 76: remove

helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
```diff
@@ -62,9 +62,11 @@ spec:
             - "--entryPoints.liveness.address=:1082"
             {{- range $i, $e := until (.Values.limits.http|int) }}
```
dtomcej, line 63: Can we use `untilStep` and use a start int of 5000 to remove the need for the adds?

https://github.com/helm/helm/issues/1055#issuecomment-552602795

integration/integration_test.go
```diff
@@ -312,7 +340,7 @@ func (s *BaseSuite) createRequiredNamespaces(c *check.C) {
 	s.kubectlCommand(c, "create", "namespace", testNamespace)
 }
 
-func (s *BaseSuite) installHelmMaesh(c *check.C, smi bool, kubeDNS bool) error {
+func (s *BaseSuite) installHelmMaesh(c *check.C, smi bool, kubeDNS bool, acl bool) error {
```
dtomcej, line 343:
```suggestion
func (s *BaseSuite) installHelmMaesh(c *check.C, kubeDNS bool, acl bool) error {
```

integration/integration_test.go
```diff
@@ -508,12 +532,171 @@ func (s *BaseSuite) digHost(c *check.C, source, namespace, destination string) {
...
+	service := config.HTTP.Services["readiness"]
+	c.Assert(service, checker.NotNil)
+
+	c.Assert(service.LoadBalancer.Servers, checker.HasLen, 1)
+	c.Assert(service.LoadBalancer.Servers[0].URL, checker.Equals, "http://127.0.0.1:8080")
+}
+
+func (s *BaseSuite) checkHTTPServiceLoadBalancer(c *check.C, config *dynamic.Configuration, svc *corev1.Service, pods []*corev1.Pod) {
+	for _, port := range svc.Spec.Ports {
+		svcKey := fmt.Sprintf("%s-%s-%d", svc.Namespace, svc.Name, port.Port)
```
dtomcej, line 585: Can't we use the provider keygen funcs to generate here?

integration/integration_test.go
```diff
@@ -508,12 +532,171 @@ func (s *BaseSuite) digHost(c *check.C, source, namespace, destination string) {
...
+			}
+
+			c.Assert(found, checker.True)
+		}
+	}
+}
+
+func (s *BaseSuite) checkTCPServiceLoadBalancer(c *check.C, config *dynamic.Configuration, svc *corev1.Service, pods []*corev1.Pod) {
+	for _, port := range svc.Spec.Ports {
+		svcKey := fmt.Sprintf("%s-%s-%d", svc.Namespace, svc.Name, port.Port)
```
dtomcej, line 612: use keygen?

integration/integration_test.go
```diff
@@ -508,12 +532,171 @@ func (s *BaseSuite) digHost(c *check.C, source, namespace, destination string) {
...
+			}
+
+			c.Assert(found, checker.True)
+		}
+	}
+}
+
+func (s *BaseSuite) checkTrafficTargetLoadBalancer(c *check.C, config *dynamic.Configuration, tt *access.TrafficTarget, svc *corev1.Service, pods []*corev1.Pod) {
+	for _, port := range svc.Spec.Ports {
+		svcKey := fmt.Sprintf("%s-%s-%s-%d-traffic-target", svc.Namespace, svc.Name, tt.Name, port.Port)
```
dtomcej, line 639: keygen?

integration/integration_test.go
```diff
@@ -508,12 +532,171 @@ func (s *BaseSuite) digHost(c *check.C, source, namespace, destination string) {
...
+				}
+			}
+
+			c.Assert(found, checker.True)
 		}
 	}
+}
+
+func (s *BaseSuite) checkTrafficTargetWhitelistDirect(c *check.C, config *dynamic.Configuration, tt *access.TrafficTarget, svc *corev1.Service, pods []*corev1.Pod) {
+	middlewareKey := fmt.Sprintf("%s-%s-%s-whitelist-traffic-target-direct", svc.Namespace, svc.Name, tt.Name)
```
dtomcej, line 665: keygen

integration/integration_test.go
```diff
@@ -508,12 +532,171 @@ func (s *BaseSuite) digHost(c *check.C, source, namespace, destination string) {
...
+
+	s.checkWhitelistSourceRange(c, config, middlewareKey, svc, pods)
+}
+
+func (s *BaseSuite) checkTrafficTargetWhitelistIndirect(c *check.C, config *dynamic.Configuration, tt *access.TrafficTarget, svc *corev1.Service, pods []*corev1.Pod) {
+	middlewareKey := fmt.Sprintf("%s-%s-%s-whitelist-traffic-target-indirect", svc.Namespace, svc.Name, tt.Name)
```
dtomcej, line 671: keygen

integration/integration_test.go
```diff
@@ -508,12 +532,171 @@ func (s *BaseSuite) digHost(c *check.C, source, namespace, destination string) {
...
 
-	return false
+	s.checkWhitelistSourceRange(c, config, middlewareKey, svc, pods)
+}
+
+func (s *BaseSuite) checkTrafficSplitWhitelistDirect(c *check.C, config *dynamic.Configuration, ts *split.TrafficSplit, svc *corev1.Service, pods []*corev1.Pod) {
+	middlewareKey := fmt.Sprintf("%s-%s-%s-whitelist-traffic-split-direct", svc.Namespace, svc.Name, ts.Name)
```
dtomcej, line 677: keygen

pkg/controller/controller.go
```diff
@@ -61,9 +60,9 @@ type Controller struct {
 	handler              *Handler
 	serviceManager       ServiceManager
 	configRefreshChan    chan string
-	provider             base.Provider
+	provider             *provider.Provider
```
dtomcej, line 63: Can we keep this an interface so that we can ensure that an interface can be used by EE for integration?

pkg/controller/controller.go
```diff
@@ -150,38 +153,36 @@ func (c *Controller) init() {
...
+	c.accessFactory = accessInformer.NewSharedInformerFactoryWithOptions(c.clients.GetAccessClient(), k8s.ResyncPeriod)
+	c.splitFactory = splitInformer.NewSharedInformerFactoryWithOptions(c.clients.GetSplitClient(), k8s.ResyncPeriod)
+	c.specsFactory = specsInformer.NewSharedInformerFactoryWithOptions(c.clients.GetSpecsClient(), k8s.ResyncPeriod)
+
+	c.accessFactory.Access().V1alpha1().TrafficTargets().Informer().AddEventHandler(c.handler)
+	c.splitFactory.Split().V1alpha2().TrafficSplits().Informer().AddEventHandler(c.handler)
+	c.specsFactory.Specs().V1alpha1().HTTPRouteGroups().Informer().AddEventHandler(c.handler)
+	c.specsFactory.Specs().V1alpha1().TCPRoutes().Informer().AddEventHandler(c.handler)
+
+	// Create the SMI listers
```
dtomcej, line 169: Can we move this to line 145 to keep the listers all together?

pkg/controller/controller.go
```diff
@@ -150,38 +153,36 @@ func (c *Controller) init() {
...
-		c.TrafficSplitLister = c.splitFactory.Split().V1alpha2().TrafficSplits().Lister()
-
-		c.provider = smi.New(c.log, c.defaultMode, c.tcpStateTable, c.ignored, c.ServiceLister, c.EndpointsLister, c.PodLister, c.TrafficTargetLister, c.HTTPRouteGroupLister, c.TCPRouteLister, c.TrafficSplitLister, c.minHTTPPort, c.maxHTTPPort)
-
-		return
-	}
-
-	// If SMI is not configured, use the kubernetes provider.
-	c.provider = kubernetes.New(c.log, c.defaultMode, c.tcpStateTable, c.ignored, c.ServiceLister, c.EndpointsLister, c.minHTTPPort, c.maxHTTPPort)
+	// Create new SharedInformerFactories, and register the event handler to informers.
```
dtomcej, line 159: Since we are changing which informers are being used at any time, we need to update the prepare check to ensure informers start properly.

pkg/controller/controller.go
```diff
@@ -150,38 +153,36 @@ func (c *Controller) init() {
...
+	c.accessFactory = accessInformer.NewSharedInformerFactoryWithOptions(c.clients.GetAccessClient(), k8s.ResyncPeriod)
+	c.splitFactory = splitInformer.NewSharedInformerFactoryWithOptions(c.clients.GetSplitClient(), k8s.ResyncPeriod)
+	c.specsFactory = specsInformer.NewSharedInformerFactoryWithOptions(c.clients.GetSpecsClient(), k8s.ResyncPeriod)
+
+	c.accessFactory.Access().V1alpha1().TrafficTargets().Informer().AddEventHandler(c.handler)
```
dtomcej, line 164: We are going to watch TrafficTargets even if ACL is disabled?

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (p *Provider) buildServicesAndRoutersForService(cfg *dynamic.Configuration, svc *topology.Service, scheme, trafficType string, middlewares []string) error {
+	switch trafficType {
+	case k8s.ServiceTypeHTTP:
+		httpRule := fmt.Sprintf("Host(`%s.%s.maesh`) || Host(`%s`)", svc.Name, svc.Namespace, svc.ClusterIP)
```
dtomcej, line 151: Isn't there a RuleGen func for this?

helm/chart/maesh/templates/controller/controller-deployment.yaml
jspdown: I considered this option but instead we chose to interpret the option and do the backward compatibility directly in the core. The solution you are proposing is compatible only when you're using Maesh via Helm. Most of our users do but iif you do it in the code you can cover Helm installs and manual installs.

helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
jspdown: Yes definitely :+1:

integration/integration_test.go
jspdown: I thought about it too. But I would prefer not to expose functions from the `key.go` just for our integration tests. WDYT?

pkg/controller/controller.go
jspdown: Oh :fearful: We completely forgot about this.

pkg/controller/controller.go
jspdown: :thinking: We should probably avoid this.

integration/integration_test.go
dtomcej: Seems ok. Just didn't want to dup code.

pkg/controller/controller.go
```diff
@@ -138,7 +137,11 @@ func NewMeshController(clients k8s.Client, cfg MeshControllerConfig) (*Controlle
 func (c *Controller) init() {
 	// Create a new SharedInformerFactory, and register the event handler to informers.
 	c.kubernetesFactory = informers.NewSharedInformerFactoryWithOptions(c.clients.GetKubernetesClient(), k8s.ResyncPeriod)
+
+	// Create the base listers
```
Ullaakut, line 141:
```suggestion
	// Create the base listers.
```

pkg/controller/controller.go
Ullaakut: And add a dot at the end of the sentence for consistency :)

pkg/controller/controller.go
```diff
@@ -150,38 +153,36 @@ func (c *Controller) init() {
...
+		c.ServiceLister,
+		c.EndpointsLister,
+		c.PodLister,
+		c.TrafficTargetLister,
+		c.TrafficSplitLister,
+		c.HTTPRouteGroupLister,
+		c.TCPRouteLister,
+		c.log)
+
+	c.provider = provider.New(topologyBuilder, c.tcpStateTable, c.ignored, c.minHTTPPort, c.maxHTTPPort, c.aclEnabled, c.defaultMode, c.meshNamespace, c.log)
```
Ullaakut, line 185: That's a lot of function parameters 👀

pkg/controller/controller.go
```diff
@@ -266,29 +268,27 @@ func (c *Controller) startInformers(stopCh <-chan struct{}, syncTimeout time.Dur
...
-		c.accessFactory.Start(stopCh)
+	c.accessFactory.Start(stopCh)
 
-		for t, ok := range c.accessFactory.WaitForCacheSync(ctx.Done()) {
-			if !ok {
-				c.log.Errorf("timed out waiting for controller caches to sync: %s", t.String())
-			}
+	for t, ok := range c.accessFactory.WaitForCacheSync(ctx.Done()) {
+		if !ok {
+			c.log.Errorf("timed out waiting for controller caches to sync: %s", t.String())
```
Ullaakut, line 275:
```suggestion
			c.log.Errorf("timed out waiting for controller cache to sync: %s", t.String())
```

pkg/controller/controller.go
```diff
@@ -266,29 +268,27 @@ func (c *Controller) startInformers(stopCh <-chan struct{}, syncTimeout time.Dur
...
-		c.specsFactory.Start(stopCh)
+	c.specsFactory.Start(stopCh)
 
-		for t, ok := range c.specsFactory.WaitForCacheSync(ctx.Done()) {
-			if !ok {
-				c.log.Errorf("timed out waiting for controller caches to sync: %s", t.String())
-			}
+	for t, ok := range c.specsFactory.WaitForCacheSync(ctx.Done()) {
+		if !ok {
+			c.log.Errorf("timed out waiting for controller caches to sync: %s", t.String())
```
Ullaakut, line 283:
```suggestion
			c.log.Errorf("timed out waiting for controller cache to sync: %s", t.String())
```

pkg/controller/controller.go
```diff
@@ -266,29 +268,27 @@ func (c *Controller) startInformers(stopCh <-chan struct{}, syncTimeout time.Dur
...
-		c.splitFactory.Start(stopCh)
+	c.splitFactory.Start(stopCh)
 
-		for t, ok := range c.splitFactory.WaitForCacheSync(ctx.Done()) {
-			if !ok {
-				c.log.Errorf("timed out waiting for controller caches to sync: %s", t.String())
-			}
+	for t, ok := range c.splitFactory.WaitForCacheSync(ctx.Done()) {
+		if !ok {
+			c.log.Errorf("timed out waiting for controller caches to sync: %s", t.String())
```
Ullaakut, line 291:
```suggestion
			c.log.Errorf("timed out waiting for controller cache to sync: %s", t.String())
```

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/containous/maesh/pkg/k8s"
+	"github.com/containous/maesh/pkg/topology"
+	"github.com/containous/traefik/v2/pkg/config/dynamic"
+)
+
+// MiddlewareBuilder is capable of building middlewares for a service.
```
Ullaakut, line 13:
```suggestion
// MiddlewareBuilder builds middlewares for a service.
```

Ullaakut: Actually, doesn't it build of middlewares _of_ a service?

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+type MiddlewareBuilder interface {
+	Build(svc *topology.Service) (*dynamic.Middleware, error)
+}
+
+// AnnotationBasedMiddlewareBuilder build middlewares from service annotations.
+type AnnotationBasedMiddlewareBuilder struct{}
```
Ullaakut, line 19: If the sole purpose of this struct is to implement the interface `MiddlewareBuilder`, I'd rather see this:

```go
type MiddlewareBuilder func(*topology.Service) (*dynamic.Middleware, error) 

func Build(svc *topology.Service) (*dynamic.Middleware, error) {
    // ...
}
```

No need for an interface and an empty struct this way :) And the MiddlewareBuilder type can still be matched by a mock function in the tests.

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// Build rate-limit middleware.
+	rateLimitAverage, hasRateLimitAverage := svc.Annotations[k8s.AnnotationRateLimitAverage]
+	rateLimitBurst, hasRateLimitBurst := svc.Annotations[k8s.AnnotationRateLimitBurst]
+
+	if hasRateLimitAverage && hasRateLimitBurst {
+		average, err := strconv.Atoi(rateLimitAverage)
+		if err != nil {
+			return nil, fmt.Errorf("unable to build rate-limit middleware, %q annotaiton is invalid: %w", k8s.AnnotationRateLimitAverage, err)
```
Ullaakut, line 47:
```suggestion
			return nil, fmt.Errorf("unable to build rate-limit middleware, %q annotation is invalid: %w", k8s.AnnotationRateLimitAverage, err)
```

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		burst, err := strconv.Atoi(rateLimitBurst)
+		if err != nil {
+			return nil, fmt.Errorf("unable to build rate-limit middleware, %q annotaiton is invalid: %w", k8s.AnnotationRateLimitBurst, err)
```
Ullaakut, line 52:
```suggestion
			return nil, fmt.Errorf("unable to build rate-limit middleware, %q annotation is invalid: %w", k8s.AnnotationRateLimitBurst, err)
```

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	if middleware.CircuitBreaker == nil && middleware.Retry == nil && middleware.RateLimit == nil {
+		return nil, nil
+	}
+
+	return &middleware, nil
+}
+
+// buildWhitelistMiddlewareFromTrafficTargetDirect builds an IPWhiteList middleware which blocks every requests except
```
Ullaakut, line 72:
```suggestion
// buildWhitelistMiddlewareFromTrafficTargetDirect builds an IPWhiteList middleware which blocks requests from
```

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+// those originating from an authorized Pod. Authorized Pods are all the Pods listed in the ServiceTrafficTarget.Sources.
```
Ullaakut, line 73:
```suggestion
// unauthorized Pods. Authorized Pods are those listed in the ServiceTrafficTarget.Sources.
```

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return &dynamic.Middleware{
+		IPWhiteList: &dynamic.IPWhiteList{
+			SourceRange: IPs,
+		},
+	}
+}
+
+// buildWhitelistMiddlewareFromTrafficSplitDirect builds an IPWhiteList middleware which blocks every requests except
```
Ullaakut, line 91:
```suggestion
// buildWhitelistMiddlewareFromTrafficSplitDirect builds an IPWhiteList middleware which blocks requests from
```

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+// those originating from an authorized Pod. Authorized pods are the pods that can access all the leaves of the
```
Ullaakut, line 92:
```suggestion
// unauthorized Pods. Authorized Pods are those that can access all the leaves of the TrafficSplit.
```

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+// TrafficSplit.
```
Ullaakut, line 93:
```suggestion
```

Was moved in the suggestion above.

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	return &dynamic.Middleware{
+		IPWhiteList: &dynamic.IPWhiteList{
+			SourceRange: IPs,
+		},
+	}
+}
+
+// buildWhitelistMiddlewareFromTrafficTargetIndirect builds an IPWhiteList middleware like
```
Ullaakut, line 108: > // buildWhitelistMiddlewareFromTrafficTargetIndirect builds an IPWhiteList middleware like
// buildWhitelistMiddlewareFromTrafficTargetDirect

You lost me there :p Could you please describe it so that it can be understood without having to look up the direct version? This comment would instantly become technical debt if the referenced function changed, is deleted, etc.

pkg/provider/middleware.go
```diff
@@ -0,0 +1,130 @@
...
+func buildWhitelistMiddlewareFromTrafficTargetIndirect(tt *topology.ServiceTrafficTarget) *dynamic.Middleware {
+	whitelist := buildWhitelistMiddlewareFromTrafficTargetDirect(tt)
+	whitelist.IPWhiteList.IPStrategy = &dynamic.IPStrategy{
+		Depth: 1,
+	}
+
+	return whitelist
+}
+
+// buildWhitelistMiddlewareFromTrafficSplitIndirect builds an IPWhiteList middleware like
```
Ullaakut, line 120: Could you please describe it so that it can be understood without having to look up the direct version? This comment would instantly become technical debt if the referenced function changed, is deleted, etc.

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+	"net"
+	"strconv"
+
+	"github.com/containous/maesh/pkg/k8s"
+	"github.com/containous/maesh/pkg/topology"
+	"github.com/containous/traefik/v2/pkg/config/dynamic"
+	"github.com/sirupsen/logrus"
+)
+
+// TopologyBuilder is capable of building Topologies.
```
Ullaakut, line 15:
```suggestion
// TopologyBuilder builds Topologies.
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+type TopologyBuilder interface {
+	Build(ignored k8s.IgnoreWrapper) (*topology.Topology, error)
```
Ullaakut, line 17: A variable called `ignored`, looks like a red flag to me! Could you add a quick comment above this method to explain why this variable is named `ignored` here?

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+}
+
+// TCPPortFinder is capable for finding a TCP port mapping.
```
Ullaakut, line 20:
```suggestion
// TCPPortFinder finds TCP port mappings.
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+type TCPPortFinder interface {
+	Find(svc k8s.ServiceWithPort) (int32, bool)
+}
+
+// When multiple Traefik Routers listen to the same entrypoint and have the same Rule, the chosen router will be the one
```
Ullaakut, line 25:
```suggestion
// When multiple Traefik Routers listen to the same entrypoint and have the same Rule, the chosen router is the one
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+// with the highest priority. There are few cases where this priority is crucial when building the dynamic configuration:
```
Ullaakut, line 26:
```suggestion
// with the highest priority. There are a few cases where this priority is crucial when building the dynamic configuration:
```

Careful, `few` and `a few` do not have the same meaning! :)

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+// - When a TrafficSplit is set on a k8s service, this will create 2 Traefik Routers. One for accessing the k8s service
```
Ullaakut, line 27:
```suggestion
// - When a TrafficSplit is set on a k8s service, 2 Traefik Routers are created. One for accessing the k8s service
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+//   endpoints and one for accessing the services endpoints mentioned in the TrafficSplit. They both have the same Rule
+//   but we should always prioritize the TrafficSplit. Therefore, TrafficSplit Routers should always have a higher priority.
+// - When a TrafficTarget Destination targets pods of a k8s service and a TrafficSplit is set on this service. This
+//   will create 2 Traefik Routers. One for the TrafficSplit and one for the TrafficTarget. We should always prioritize
```
Ullaakut, line 31:
```suggestion
//   creates 2 Traefik Routers. One for the TrafficSplit and one for the TrafficTarget. We should always prioritize
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+//   TrafficSplits Routers and so, TrafficSplit Routers should always have a higher priority than TrafficTarget Routers.
```
Ullaakut, line 32:
```suggestion
//   TrafficSplits Routers and TrafficSplit Routers should always have a higher priority than TrafficTarget Routers.
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+const (
+	priorityService               = 1
```
Ullaakut, line 34: Please use `iota` here.

https://programming.guide/go/iota.html

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+	topologyBuilder   TopologyBuilder
+	tcpStateTable     TCPPortFinder
+	middlewareBuilder MiddlewareBuilder
+	ignored           k8s.IgnoreWrapper
+
+	logger logrus.FieldLogger
+}
+
+// New creates a new Provider.
+func New(topologyBuilder TopologyBuilder, tcpStateTable TCPPortFinder, ignored k8s.IgnoreWrapper, minHTTPPort, maxHTTPPort int32, acl bool, defaultTrafficType, maeshNamespace string, logger logrus.FieldLogger) *Provider {
```
Ullaakut, line 57: At this point, WDYT of building a structure containing the dependencies of the Provider, so that we can have only one argument here and make the use of this `New` function more readable?

(I'm not saying we need variadic modifiers for building the provider, just a struct that we would build explicitly before calling new, which would make it clearer how it works, IMO)

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+	acl                bool
+	minHTTPPort        int32
+	maxHTTPPort        int32
+	defaultTrafficType string
+	maeshNamespace     string
+
+	topologyBuilder   TopologyBuilder
+	tcpStateTable     TCPPortFinder
+	middlewareBuilder MiddlewareBuilder
+	ignored           k8s.IgnoreWrapper
```
Ullaakut, line 51: I'm still not sure what this is for, could you add a comment here as well?

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+
+			svcKey := getServiceKeyFromTrafficTarget(tt, svcPort.Port)
+			cfg.HTTP.Services[svcKey] = buildHTTPServiceFromTrafficTarget(tt, scheme, svcPort.TargetPort.IntVal)
+
+			rtrMiddlewares := addToSliceCopy(middlewares, whitelistDirectKey)
+
+			directRtrKey := getRouterKeyFromTrafficTargetDirect(tt, svcPort.Port)
+			cfg.HTTP.Routers[directRtrKey] = buildHTTPRouter(rule, entrypoint, rtrMiddlewares, svcKey, priorityTrafficTargetDirect)
+
+			// If the ServiceTrafficTarget is a backend of at least one TrafficSplit we need an additional router with
```
Ullaakut, line 210:
```suggestion
			// If the ServiceTrafficTarget is the backend of at least one TrafficSplit we need an additional router with
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+	return nil
+}
+
+func (p *Provider) buildBlockAllRouters(cfg *dynamic.Configuration, svc *topology.Service) {
+	rule := buildHTTPRuleFromService(svc)
+
+	for portID, svcPort := range svc.Ports {
+		entrypoint, err := p.buildHTTPEntrypoint(portID)
+		if err != nil {
+			p.logger.Errorf("unable to build HTTP entrypoint for Service %s/%s and portID %d: %w", svc.Namespace, svc.Name, portID, err)
```
Ullaakut, line 347:
```suggestion
			p.logger.Errorf("unable to build HTTP entrypoint for Service %s/%s and port %d: %w", svc.Namespace, svc.Name, portID, err)
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+				cfg.HTTP.Services[backendSvcKey] = buildHTTPSplitTrafficBackendService(backend, scheme, svcPort.Port)
+				backendSvcs[i] = dynamic.WRRService{
+					Name:   backendSvcKey,
+					Weight: getIntRef(backend.Weight),
+				}
+			}
+
+			entrypoint, err := p.buildHTTPEntrypoint(portID)
+			if err != nil {
+				p.logger.Errorf("Unable to build HTTP entrypoint for Service %s/%s and portID %d: %v", ts.Service.Namespace, ts.Service.Name, portID, err)
```
Ullaakut, line 284:
```suggestion
				p.logger.Errorf("Unable to build HTTP entrypoint for Service %s/%s and port %d: %v", ts.Service.Namespace, ts.Service.Name, portID, err)
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+		whitelistDirect := buildWhitelistMiddlewareFromTrafficTargetDirect(tt)
+		whitelistDirectKey := getWhitelistMiddlewareKeyFromTrafficTargetDirect(tt)
+		cfg.HTTP.Middlewares[whitelistDirectKey] = whitelistDirect
+
+		rule := buildHTTPRuleFromTrafficTarget(tt)
+
+		for portID, svcPort := range tt.Destination.Ports {
+			entrypoint, err := p.buildHTTPEntrypoint(portID)
+			if err != nil {
+				p.logger.Errorf("Unable to build HTTP entrypoint for Service %s/%s and portID %d: %v", tt.Service.Namespace, tt.Service.Name, portID, err)
```
Ullaakut, line 198:
```suggestion
				p.logger.Errorf("Unable to build HTTP entrypoint for Service %s/%s and port %d: %v", tt.Service.Namespace, tt.Service.Name, portID, err)
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+
+func (p *Provider) buildServicesAndRoutersForService(cfg *dynamic.Configuration, svc *topology.Service, scheme, trafficType string, middlewares []string) error {
+	switch trafficType {
+	case k8s.ServiceTypeHTTP:
+		httpRule := fmt.Sprintf("Host(`%s.%s.maesh`) || Host(`%s`)", svc.Name, svc.Namespace, svc.ClusterIP)
+
+		for portID, svcPort := range svc.Ports {
+			entrypoint, err := p.buildHTTPEntrypoint(portID)
+			if err != nil {
+				p.logger.Errorf("Unable to build HTTP entrypoint for Service %s/%s and portID %d: %v", svc.Namespace, svc.Name, portID, err)
```
Ullaakut, line 156:
```suggestion
				p.logger.Errorf("Unable to build HTTP entrypoint for Service %s/%s and port %d: %v", svc.Namespace, svc.Name, portID, err)
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+			Rule:        rule,
+			Priority:    priorityService,
+		}
+	}
+}
+
+func (p Provider) buildHTTPEntrypoint(portID int) (string, error) {
+	port := p.minHTTPPort + int32(portID)
+	if port >= p.maxHTTPPort {
+		return "", errors.New("too many HTTP entrypoint")
```
Ullaakut, line 365:
```suggestion
		return "", errors.New("too many HTTP entrypoints")
```

pkg/provider/provider.go
```diff
@@ -0,0 +1,604 @@
...
+
+func (p *Provider) getTrafficTypeAnnotation(svc *topology.Service) (string, error) {
+	trafficType, ok := svc.Annotations[k8s.AnnotationServiceType]
+
+	if !ok {
+		return p.defaultTrafficType, nil
+	}
+
+	if trafficType != k8s.ServiceTypeHTTP && trafficType != k8s.ServiceTypeTCP {
+		return "", fmt.Errorf("traffic-type annotation references an unknown traffic type %q", trafficType)
```
Ullaakut, line 393:
```suggestion
		return "", fmt.Errorf("traffic-type annotation references an unsupported traffic type %q", trafficType)
```

pkg/provider/provider_test.go
```diff
@@ -0,0 +1,839 @@
...
+	spec "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/specs/v1alpha1"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+type TopologyBuilderMock func() (*topology.Topology, error)
```
Ullaakut, line 20: Should not be exported

pkg/provider/provider_test.go
```diff
@@ -0,0 +1,839 @@
...
+
+func (m TopologyBuilderMock) Build(_ mk8s.IgnoreWrapper) (*topology.Topology, error) {
```
Ullaakut, line 22: Can be a simple function if instead of having a `MiddlewareBuilder` interface we just define a MiddlewareBuilder as a `func(mk8s.IgnoreWrapper) (*topology.Topology, error)`

pkg/provider/provider_test.go
```diff
@@ -0,0 +1,839 @@
...
+				},
+			},
+		},
+	}
+
+	assert.Equal(t, want, got)
+}
+
+func TestProvider_BuildConfigTCP(t *testing.T) {
+	saA := "sa-a"
```
Ullaakut, line 218: These tests are very difficult to review/maintain 😅 I'm not sure what to propose to make it better while keeping the same depth in testing though! 🤔

pkg/topology/builder.go
```diff
@@ -0,0 +1,667 @@
...
+	trafficTargetLister  accessLister.TrafficTargetLister
+	trafficSplitLister   splitLister.TrafficSplitLister
+	httpRouteGroupLister specLister.HTTPRouteGroupLister
+	tcpRoutesLister      specLister.TCPRouteLister
+	logger               logrus.FieldLogger
+}
+
+// NewBuilder creates a new Builder.
+func NewBuilder(
+	svcLister listers.ServiceLister,
```
Ullaakut, line 35: Once again, that's a lot of parameters. I'd suggest using a structure here too.

pkg/topology/builder.go
```diff
@@ -0,0 +1,667 @@
...
+		svcKey := NameNamespace{svc.Name, svc.Namespace}
+
+		svc, ok := topology.Services[svcKey]
+		if !ok {
+			return fmt.Errorf("unable to find Service %s/%s", svc.Namespace, svc.Name)
+		}
+
+		var destPods []*Pod
+
+		// Find out who are the destination pods.
```
Ullaakut, line 136:
```suggestion
		// Find out which are the destination pods.
```

pkg/topology/builder.go
```diff
@@ -0,0 +1,667 @@
...
+		for _, pod := range pods {
+			if pod.Status.PodIP == "" {
+				continue
+			}
+
+			destPods = append(destPods, getOrCreatePod(topology, pod))
+		}
+
+		// Find out which port can be used on the destination service.
```
Ullaakut, line 145:
```suggestion
		// Find out which ports can be used on the destination service.
```

Right?

pkg/topology/builder.go
```diff
@@ -0,0 +1,667 @@
...
+		svc.TrafficTargets = append(svc.TrafficTargets, svcTT)
+
+		// Add the ServiceTrafficTarget to the source pods.
+		for _, source := range sources {
+			for _, pod := range source.Pods {
+				pod.Outgoing = append(pod.Outgoing, svcTT)
+			}
+		}
+
+		// Add the ServiceTrafficTarget to the destination pods
```
Ullaakut, line 173:
```suggestion
		// Add the ServiceTrafficTarget to the destination pods.
```

pkg/topology/builder.go
```diff
@@ -0,0 +1,667 @@
...
+
+// populateTrafficSplitsAuthorizedIncomingTraffic computes the list of pods allowed to access a traffic-split. As
+// traffic-splits may form a graph, it has to be done once all the traffic-splits have been processed. To avoid runtime
+// issues, this method detects cycles in the graph.
+func (b *Builder) populateTrafficSplitsAuthorizedIncomingTraffic(topology *Topology) {
+	loopDetected := make(map[*Service][]*TrafficSplit)
+
+	for _, svc := range topology.Services {
+		for _, ts := range svc.TrafficSplits {
+			pods, err := b.getIncomingPodsForTrafficSplit(ts, map[NameNamespace]bool{})
```
Ullaakut, line 247: Why are you passing a new empty map to the function instead of creating it at the beginning of the func? 🤔 If you were reusing it between function calls I'd understand but this seems like a mistake.

Also, when storing such information as whether a name+namespace combo has been visited already or not, since you will never store `false`, using a `bool` here is overkill. Using a `struct{}` is actually more explicit here.

([Brought to you by Dave](https://dave.cheney.net/2014/03/25/the-empty-struct))

pkg/topology/builder.go
```diff
@@ -0,0 +1,667 @@
...
+	}
+
+	return TrafficSpec{
+		TCPRoute: tcpRoute,
+	}, nil
+}
+
+// getTrafficTargetDestinationPorts gets the ports mentioned in the TrafficTarget.Destination.Port.
+// If the port is "", it will returns all the ports of the Service.
+// If the port is an integer, it will returns on this port.
```
Ullaakut, line 464:
```suggestion
// If the port is an integer, it is returned.
```

pkg/topology/builder.go
```diff
@@ -0,0 +1,667 @@
...
+		return TrafficSpec{}, fmt.Errorf("unable to find TCPRoute %s/%s", ns, s.Name)
+	}
+
+	return TrafficSpec{
+		TCPRoute: tcpRoute,
+	}, nil
+}
+
+// getTrafficTargetDestinationPorts gets the ports mentioned in the TrafficTarget.Destination.Port.
+// If the port is "", it will returns all the ports of the Service.
```
Ullaakut, line 463:
```suggestion
// If the port is "", all of the Service's ports are returned.
```

pkg/topology/builder.go
```diff
@@ -0,0 +1,667 @@
...
+}
+
+type resources struct {
+	Services        map[NameNamespace]*v1.Service
+	TrafficTargets  map[NameNamespace]*access.TrafficTarget
+	TrafficSplits   map[NameNamespace]*split.TrafficSplit
+	HTTPRouteGroups map[NameNamespace]*spec.HTTPRouteGroup
+	TCPRoutes       map[NameNamespace]*spec.TCPRoute
+
+	// Pods indexes
```
Ullaakut, line 507:
```suggestion
	// Pods indexes.
```

pkg/topology/builder.go
```diff
@@ -0,0 +1,667 @@
...
+		res.Services[NameNamespace{svc.Name, svc.Namespace}] = svc
+	}
+
+	b.indexSMIResources(res, ignored, tts, tss, tcpRts, httpRtGrps)
+	b.indexPods(res, ignored, pods, eps)
+
+	return res, nil
+}
+
+// indexPods populate the different pod indexes in the given resources object. It builds 3 indexes:
```
Ullaakut, line 574:
```suggestion
// indexPods populates the different pod indexes in the given resources object. It builds 3 indexes:
```

pkg/topology/topology.go
```diff
@@ -0,0 +1,116 @@
+package topology
+
+import (
+	specs "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/specs/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NameNamespace is a key for referencing unique resources.
+type NameNamespace struct {
```
Ullaakut, line 10: Shouldn't it be called a `topology.Key` then? I'm not fond of having a key called by what it contains 🤔

Ullaakut: Also this would make it more resilient to change (unlikely to happen but hey) :)

pkg/topology/topology.go
```diff
@@ -0,0 +1,116 @@
...
+	Namespace string
+}
+
+// Topology holds the graph. Each Pods and services are nodes of the graph.
+type Topology struct {
+	Services map[NameNamespace]*Service
+	Pods     map[NameNamespace]*Pod
+}
+
+// NewTopology create a new Topology.
```
Ullaakut, line 21:
```suggestion
// NewTopology creates a new Topology.
```

pkg/topology/topology.go
```diff
@@ -0,0 +1,116 @@
...
+	// List of TrafficTargets that are targeting pods which are selected by this service.
+	TrafficTargets []*ServiceTrafficTarget
+	// List of TrafficSplits that are targeting this service.
+	TrafficSplits []*TrafficSplit
+	// List of TrafficSplit mentioning this service as a backend.
+	BackendOf []*TrafficSplit
+}
+
+// ServiceTrafficTarget represents a TrafficTarget applied a on Service. TrafficTargets have a Destination service
+// account. This service account can be set on many pods, each of them, potentially accessible via different services.
```
Ullaakut, line 48:
```suggestion
// account. This service account can be set on many pods, each of them, potentially accessible through different services.
```

integration/integration_test.go
```diff
@@ -53,14 +55,15 @@ func Test(t *testing.T) {
 		return
 	}
 
+	check.Suite(&ACLDisabledSuite{})
```
dtomcej, line 58: Since you modified the suites, you should probably also update `.semaphore/semaphore.yaml` so that they run in the CI 😝

pkg/provider/provider.go
jspdown: We didn't think about it too much since it was named almost everywhere else. But that's right, `ignoredResources` would make more sense.

pkg/provider/provider.go
jspdown: I kind of agree with you. Putting everything related to Maesh configuration in a proper struct would make more sense. So I would put:
- ignored k8s.IgnoreWrapper
- minHTTPPort
- maxHTTPPort int32
- acl bool
- defaultTrafficType string
- maeshNamespace string

pkg/provider/provider.go
jspdown: We should probably rename `portID 5` into `port number 5` but definitely not `port`. This value represents the ID of the port in the list of ports.

pkg/topology/builder.go
jspdown: I agree for the `struct{}` :+1: 
The `map[NameNamespace]struct{}` represents the visited nodes when traversing the graph to detect cycles. It must be a new map each time we call this function because we are interested in detecting cycles on the given TrafficSplit.

integration/integration_test.go
jspdown: Actually, I think we should keep this parameter and add a new test which checks if the `--smi` option is still handled correctly.

docs/content/examples.md
```diff
@@ -243,9 +243,9 @@ Accept: */*
 X-Forwarded-For: 3.4.5.6
 ```
 
-## SMI Example
+## ACL Example
 
-In order to expose the HTTP Services from above with SMI enabled, please create the following resources:
+In order to expose the HTTP Services from above with ACL enabled, please create the following resources:
```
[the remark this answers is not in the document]
jspdown: I suppose it refers to the `Simple Example` section, the example above.

integration/acl_disabled_test.go
```diff
@@ -0,0 +1,102 @@
+package integration
+
+import (
+	"github.com/go-check/check"
+	checker "github.com/vdemeester/shakers"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// ACLDisabledSuite
```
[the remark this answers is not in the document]
jspdown: There's no purpose but since every integration test does that I just followed the rule.

integration/acl_enabled_test.go
```diff
@@ -0,0 +1,96 @@
+package integration
+
+import (
+	"github.com/go-check/check"
+	checker "github.com/vdemeester/shakers"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// ACLEnabledSuite
```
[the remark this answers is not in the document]
jspdown: Same reason

integration/integration_test.go
```diff
@@ -121,7 +124,7 @@ func (s *BaseSuite) maeshPrepareWithArgs(args ...string) *exec.Cmd {
 	return exec.Command(maeshBinary, args...)
 }
 
-func (s *BaseSuite) startMaeshBinaryCmd(c *check.C, smi bool) *exec.Cmd {
+func (s *BaseSuite) startMaeshBinaryCmd(c *check.C, smi bool, acl bool) *exec.Cmd {
```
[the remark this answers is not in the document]
jspdown: Since `smi` is still supported, we wanted to keep it in params so we can test we are not breaking compatibility.

integration/integration_test.go
```diff
@@ -326,6 +354,10 @@ func (s *BaseSuite) installHelmMaesh(c *check.C, smi bool, kubeDNS bool) error {
 		argSlice = append(argSlice, "--set", "kubedns=true")
 	}
 
+	if acl {
```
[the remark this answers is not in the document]
jspdown: As explained above, we would like to keep it and use the `smi` param to check we are not breaking compatibility.

integration/integration_test.go
SantoDE: Yes, I'd prefer that

pkg/controller/controller.go
SantoDE: +1

docs/content/configuration.md
```diff
@@ -14,22 +14,32 @@ The static configuration is configured when the Maesh service mesh is installed,
...
+    Please see the [SMI Specification](https://github.com/servicemeshinterface/smi-spec/blob/master/traffic-access-control.md) for more information.
 
 ## Dynamic configuration
 
-Dynamic configuration can be provided to Maesh using either annotations on kubernetes services (default mode) or SMI resources if Maesh is installed with [SMI enabled](./install.md#service-mesh-interface).
+Dynamic configuration can be provided to Maesh using annotations on Kubernetes services and via SMI objects. 
 
-### With Kubernetes Services
+ | Features              | ACL disabled | ACL enabled |
+ |-----------------------|--------------|-------------|
```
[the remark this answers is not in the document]
jspdown: I'm not sure to understand what you mean. It gets rendered correctly and in the source file I don't see any missing dash.

integration/integration_test.go
```diff
@@ -121,7 +124,7 @@ func (s *BaseSuite) maeshPrepareWithArgs(args ...string) *exec.Cmd {
 	return exec.Command(maeshBinary, args...)
 }
 
-func (s *BaseSuite) startMaeshBinaryCmd(c *check.C, smi bool) *exec.Cmd {
+func (s *BaseSuite) startMaeshBinaryCmd(c *check.C, smi bool, acl bool) *exec.Cmd {
 	args := []string{}
 	if smi {
```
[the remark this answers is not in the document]
jspdown: Same reason

docs/content/install.md
```diff
@@ -58,29 +58,31 @@ If you use a cluster domain other than `cluster.local` set it by using the `clus
 helm install maesh maesh/maesh --set clusterDomain=my.custom.domain.com
 ```
 
-## Service Mesh Interface
+## Access Control List
```
SantoDE, line 61: Should we maybe add, that it was used to be called "SMI" before?

docs/content/index.md
```diff
@@ -19,7 +19,7 @@ Moreover, Maesh is opt-in by default, which means that your existing services ar
 Maesh does not use any sidecar container but handles routing through proxy endpoints running on each node.
```
SantoDE, line 19: Generel question:

should we add a "deprecates" section or something like that?

integration/integration_test.go
jspdown: Done

.golangci.toml
```diff
@@ -40,6 +40,8 @@
     "funlen",
     "bodyclose", # Too many false positives
     "godox",
+    "unparam", # Often leads to bad practice
```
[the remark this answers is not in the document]
jspdown: > What case did you had in mind when claiming it leads to bad practice?
We have some functions in the tests that accept a name and a namespace. It appears that the namespace was always the same. This linter forced us to remove the namespace parameter because it was always the same. 

Also, if you intend to write a function which accepts a param which is not yet used but will be later, you can't. 

Anyway, we've reverted the change.

pkg/provider/provider_test.go
jspdown: Done, I replace the `MiddlewareBuilder` by a function.

pkg/provider/provider.go
jspdown: I will do what you suggests but instead of using the `portID` variable, I will use the `svcPort.Port`. `portID` is really an implementation details that woudn't bring a useful information to the user.

pkg/provider/provider_test.go
jspdown: Yes, I'm worried about that too. In order to test the provider we need to generate a graph, which take space and is difficult to follow. Then we have to compare the resulting dynamic config to the expected one. 

We could eventually marshal the expected dynamic config and put it in a JSON file. It could remove a lot of code, but I'm not sure it would make it more maintainable.
Then, for the topology, we could have a serializer and deserializer, and put them in a JSON file.

WDYT?

docs/content/install.md
jspdown: I don't know, It's not really the same thing. How would you write it?

docs/content/index.md
jspdown: I think we could document deprecated flags in the `install.md`, where they used to be mentioned. 
Something like this
```
# SMI Mode (Deprecated)

Maesh supports the SMI specification which defines a set of custom resources to provide a fine-grained control over instrumentation, routing and access control of east-west communications.

To enable SMI, install maesh in SMI mode by setting the smi.enable helm chart option to true.

helm install maesh --namespace=maesh maesh/maesh --set smi.enable=true`
The smi.enable option makes Maesh process SMI resources.

!!! Note This option is deprecated. Please consider using `--acl` instead.
```

.golangci.toml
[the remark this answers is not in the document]
Ullaakut: It was complaining even though you passed the namespace as a `_`? That shouldn't happen 🤔

docs/content/examples.md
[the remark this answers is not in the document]
Ullaakut: It's not super clear that it refers to that IMO

docs/content/install.md
```diff
@@ -58,29 +58,31 @@ If you use a cluster domain other than `cluster.local` set it by using the `clus
...
 ```
 
-- The `smi.enable` option makes Maesh process SMI resources.
+Maesh supports the [SMI specification](https://smi-spec.io/) which defines a set of custom resources
+to provide a fine-grained control over instrumentation, routing and access control of east-west communications.
 
 !!! Note CRDs
     Helm v3 automatically will install the CRDs in the `/crds` directory.
-    If you are re-installing into a cluster with the CRDs already present, helm may give you a warning.
+    If you are re-installing into a cluster with the CRDs already present, Helm may give you a warning.
```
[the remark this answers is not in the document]
Ullaakut: Reopening since the second point was not answered :p

docs/content/quickstart.md
```diff
@@ -1,7 +1,7 @@
 # Quickstart
 
 Maesh can be installed in your cluster without affecting any running services.
-It will safely install itself via the helm chart, and will be ready for use immediately after.
+It will safely install itself via the Helm Chart, and will be ready for use immediately after.
```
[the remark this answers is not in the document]
Ullaakut: Reopening since the second point was not answered

integration/acl_disabled_test.go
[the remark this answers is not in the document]
Ullaakut: 🤷

pkg/provider/middleware.go
Ullaakut: Reopening since second comment was left out

pkg/provider/provider_test.go
Ullaakut: Sounds good to me, might be easier to understand indeed :)

pkg/topology/builder.go
Ullaakut: It can be a new map each time by creating it at the beginning of `getIncomingPodsForTrafficSplit` instead of always passing an empty map as a parameter, is what I meant! Unless there is a case where `getIncomingPodsForTrafficSplit` does not take an empty map?

pkg/topology/builder.go
```diff
@@ -0,0 +1,652 @@
...
+}
+
+func unionPod(pods1, pods2 []*Pod) []*Pod {
+	var union []*Pod
+
+	if pods2 == nil {
+		return pods1
+	}
+
+	p := map[*Pod]bool{}
```
jlevesy, line 312:
```suggestion
	p := map[*Pod]struct{}
```

Also, using pointer as keys sounds scary and induces limitations to this function (the same pod can be represented multiple times in memory).

WDYT of adding a comment explaining why it is valid to use pointers as keys here, clearly mentioning that it should not be used outside of the context of an informer ?

pkg/topology/builder.go
```diff
@@ -0,0 +1,652 @@
...
+
+// buildTrafficTargetSources retrieves the Pod IPs for each Pod mentioned in a source of the given TrafficTarget.
+// If a Pod IP is not yet available, the pod will be skipped.
+func (b *Builder) buildTrafficTargetSources(res *resources, t *Topology, tt *access.TrafficTarget) []ServiceTrafficTargetSource {
+	sources := make([]ServiceTrafficTargetSource, len(tt.Sources))
+
+	for i, source := range tt.Sources {
+		srcSaKey := Key{source.Name, source.Namespace}
+
+		pods := res.PodsBySa[srcSaKey]
```
jlevesy, line 335:
```suggestion
		pods := res.PodsByServiceAccounts[srcSaKey]
```

I think this will help :)

pkg/topology/builder.go
```diff
@@ -0,0 +1,652 @@
...
+				}
+
+				res.PodsBySvcBySa[keySA][keyEP] = append(res.PodsBySvcBySa[keySA][keyEP], pod)
+				res.PodsBySvc[keyEP] = append(res.PodsBySvc[keyEP], pod)
+			}
+		}
+	}
+}
+
+func (b *Builder) indexSMIResources(res *resources, ignoredResources mk8s.IgnoreWrapper, tts []*access.TrafficTarget, tss []*split.TrafficSplit, tcpRts []*spec.TCPRoute, httpRtGrps []*spec.HTTPRouteGroup) {
```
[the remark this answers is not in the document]
Ullaakut: Careful with that double space though 🦅

Ullaakut:
@jspdown From my side, there are still 6 issues remaining until an LGTM from me :) You might need to expand my reviews to see the comments

docs/content/install.md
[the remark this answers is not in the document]
jspdown: Sorry, I missed the second point. I actually don't know. @dtomcej have you tried it before?

docs/content/quickstart.md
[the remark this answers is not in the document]
jspdown: I agree. We should have a link here pointing to https://helm.sh/docs/intro/using_helm/#helm-install-installing-a-package

pkg/provider/middleware.go
jspdown: Yes, it does. Will replace `for a` by `of a`.

pkg/topology/builder.go
jspdown: ``getIncomingPodsForTrafficSplit` is called by `populateTrafficSplitsAuthorizedIncomingTraffic` and the `visited` map has to be a new one then, `getIncomingPodsForTrafficSplit` is also called recursively in `getIncomingPodsForService`. This second must not create a new `visited` map as we are interested in previously visited nodes to detect cycles.

pkg/controller/controller.go
```diff
@@ -61,9 +60,9 @@ type Controller struct {
 	handler              *Handler
 	serviceManager       ServiceManager
 	configRefreshChan    chan string
-	provider             base.Provider
+	provider             *provider.Provider
 	ignored              k8s.IgnoreWrapper
-	smiEnabled           bool
+	aclEnabled           bool
```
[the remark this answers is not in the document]
jspdown: Yes, we might want to consider that. I think it should be done separately in another PR. @dtomcej WDYT?

pkg/topology/builder.go
```diff
@@ -0,0 +1,652 @@
...
+
+	return trafficSpecs, nil
+}
+
+func (b *Builder) buildHTTPRouteGroup(httpRtGrps map[Key]*spec.HTTPRouteGroup, ns string, s access.TrafficTargetSpec) (TrafficSpec, error) {
+	key := Key{s.Name, ns}
+
+	httpRouteGroup, ok := httpRtGrps[key]
+	if !ok {
+		return TrafficSpec{}, fmt.Errorf("unable to find HTTPRouteGroup %s/%s", ns, s.Name)
```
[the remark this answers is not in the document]
jspdown: The only benefit I see it that it make error cases more cleaner. If I change the returned value into a pointer, I will have to change it's called return type into a `[]*TrafficSpec` or dereference the pointer. 
I would rather prefer to keep it like that.

pkg/topology/builder.go
jspdown: Good catch! I'm going to replace the `*Pod` with a `Key`. I think we can afford the overhead since it will prevent bugs that would be really had to diagnose. 
The `*Pod` limits no only the use of this function but also it's callers.

pkg/provider/provider_test.go
jspdown: Instead of providing a serializer for the Topology, I'm seriously considering removing the circular references stuff in favour of an "index by Key" model, like in Traefik's dynamic config. This would make the structure a little less convenient to use but much more resilient and extendable.

pkg/topology/builder.go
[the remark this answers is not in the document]
jlevesy: Yeah agreed, as you wish :)

pkg/provider/provider_test.go
jspdown: Done in 7cc784e
It makes tests much more maintainable. I also took the opportunity to add more tests.

docs/content/install.md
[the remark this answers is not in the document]
jspdown: I did the test and it doesn't seem to print a warning.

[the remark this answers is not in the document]
Ullaakut: Maybe let's remove that sentence then? 🤷

PR #491, opened 2020-04-08
Add semaphore build cache

### LINKED ISSUE #481: Leverage Semaphore Build Cache
#### Proposal

Right now, every build job in semaphore downloads the pre-reqs and builds the app before testing.

It would be better to have a build step that caches the result for use in the integration tests, so that they can run without requiring a rebuild, so if they are not run totally in parallel, that we aren't wasting time or resources.

#### Background

If there are two or more builds going at once, we may only get one job running at a time, which takes much longer, and we have multiple jobs doing the exact same thing.

kevinpollet:
### What does this PR do?

Fixes #481.

Currently, every job downloads Go dependencies and build Maesh. This PR uses the available [Semaphore CI cache](https://docs.semaphoreci.com/essentials/caching-dependencies-and-directories/#cache-restore-keysecond-key) to avoid unnecessary work.

Note: `helm` and `k3d` dependencies are not cached. To cache these dependencies, we have to use a [custom container](https://docs.semaphoreci.com/ci-cd-environment/custom-ci-cd-environment-with-docker/#building-a-minimal-docker-image-for-semaphore) which seems a bit unnecessary (`helm` and `k3d` installations seem to be fast).

WDYT?

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -44,6 +44,7 @@ blocks:
         commands:
           - sem-version go 1.14
           - checkout
+          - cache restore
       jobs:
         - name: Unit Tests
           commands:
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -24,6 +24,10 @@ blocks:
          - sem-version go 1.14
          - checkout
       jobs:
+        - name: Cache Dependencies
+          commands:
+            - go mod download
+            - cache store
         - name: Build and Check
           commands:
             - make check build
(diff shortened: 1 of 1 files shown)


dtomcej:
I think that this is a great step forward.

Is there any way to not re-download the docker images between blocks?

That is the only thing that seems to be duplicated now.

kevinpollet:
It's not possible to reuse downloaded Docker images between blocks and jobs because each job runs in a fresh VM. Maybe we could use the same trick used to share the Maesh image, but we will have to maintain a Docker image dependencies list used by the integration tests which is not worth.

BTW, in the documentation the recommended way to share Docker images is the following:

> Each job runs in a fully isolated environment. This means that Docker images are not shared among jobs, not even on the same block. You should use a registry to pass images between jobs.

I've also pushed some little changes to the PR 😅

PR #495, opened 2020-04-08
Create Maesh Proxy command

### LINKED ISSUE #492: Add Proxy functionality via Traefik lib
#### Proposal

We should have our own proxy command that allows us to run Traefik via library.

#### Background

Running our own Traefik instance can give us a lot of flexibility and customization options

dtomcej:
This PR:

- Creates a proxy command based on the Traefik Library
- Uses the maesh proxy command in the maesh docker image for proxy nodes instead of traefik docker images
- Slightly modifies the setupCommand to pass our linting requirements

Fixes #492


helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
```diff
@@ -55,9 +55,10 @@ spec:
       {{- end }}
       containers:
         - name: maesh-mesh
-          image: {{ include "maesh.meshImage" . | quote }} 
-          imagePullPolicy: {{ .Values.mesh.image.pullPolicy | default "IfNotPresent" }}
+          image: {{ include "maesh.controllerImage" . | quote }}
```
kevinpollet, line 58: Maybe the `maesh.controller.image` property could be renamed to `maesh.image` as it is now used for the controller and proxy? The `maesh.controller.image` should be deprecated and still usable though.

WDYT?

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+	"github.com/containous/traefik/v2/pkg/types"
+	"github.com/containous/traefik/v2/pkg/version"
+	"github.com/coreos/go-systemd/daemon"
+	"github.com/sirupsen/logrus"
+	"github.com/vulcand/oxy/roundrobin"
+)
+
+// NewCmd builds a new Proxy command.
+func NewCmd(loaders []cli.ResourceLoader) *cli.Command {
+	// traefik config inits
```
jspdown, line 40:
```suggestion
	// Traefik config inits.
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+	sent, err := daemon.SdNotify(false, "READY=1")
+	if !sent && err != nil {
+		log.WithoutContext().Errorf("Failed to notify: %v", err)
+	}
+
+	t, err := daemon.SdWatchdogEnabled(false)
+	if err != nil {
+		log.WithoutContext().Errorf("Could not enable Watchdog: %v", err)
+	} else if t != 0 {
+		// Send a ping each half time given
```
jspdown, line 104:
```suggestion
		// Send a ping each half time given.
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+	svr.Wait()
+	log.WithoutContext().Info("Shutting down")
+
+	return nil
+}
+
+func setupServer(staticConfiguration *static.Configuration) (*server.Server, error) {
+	providerAggregator := aggregator.NewProviderAggregator(*staticConfiguration.Providers)
+
+	// adds internal provider
```
jspdown, line 135:
```suggestion
	// Adds internal provider.
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+	err := providerAggregator.AddProvider(traefik.New(*staticConfiguration))
+	if err != nil {
+		return nil, err
+	}
+
+	tlsManager := traefiktls.NewManager()
+
+	acmeProviders := initACMEProvider(staticConfiguration, &providerAggregator, tlsManager)
```
jspdown, line 143: We can probably remove ACME providers as this is not used by Maesh

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+
+	serverEntryPointsTCP, err := server.NewTCPEntryPoints(staticConfiguration.EntryPoints)
```
jspdown, line 145: I didn't thought about it when I fist saw your PR, but this is going to open the door to dynamic allocation of entrypoints! Currently, when you initialize Maesh you have to provide the maximum number of HTTP and TCP entrypoints, as your cluster grows, there will be less and less entrypoint available, until you have no entrypoint left. This is especially annoying for TCP as you needs as many TCP entrypoints as you have TCP Services in your cluster.
With this new Proxy using the Traefik library, we will be able to potentially derive the number entrypoints needed from the dynamic configuration :)

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+	if err != nil {
+		log.WithoutContext().Warnf("Unable to create access logger : %v", err)
+		return nil
+	}
+
+	return accessLoggerMiddleware
+}
+
+func configureLogging(staticConfiguration *static.Configuration) {
+	// configure default log flags
```
jspdown, line 342:
```suggestion
	// Configure default log flags.
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+	stdlog.SetFlags(stdlog.Lshortfile | stdlog.LstdFlags)
+
+	// configure log level
```
jspdown, line 345:
```suggestion
	// Configure log level.
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+	// an explicitly defined log level always has precedence. if none is
```
jspdown, line 346:
```suggestion
	// An explicitly defined log level always has precedence. If none is
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	log.SetLevel(level)
+
+	var logFile string
+	if staticConfiguration.Log != nil && len(staticConfiguration.Log.FilePath) > 0 {
+		logFile = staticConfiguration.Log.FilePath
+	}
+
+	// configure log format
```
jspdown, line 366:
```suggestion
	// Configure log format.
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+			}
+		})
+
+		if err != nil {
+			log.WithoutContext().Errorf("Error while opening log file %s: %v", logFile, err)
+		}
+	}
+}
+
+func stats(staticConfiguration *static.Configuration) {
```
jspdown, line 398: `collect.Collect` sends the static configuration to an hard coded endpoint. I think we should consider having our own endpoint for Maesh. Every Maesh users will use the same provider in their static cfg. This will deform the reality.

So I suggest we update the Makefile and add a `-X` to change the stat collection endpoint.
https://github.com/containous/traefik/blob/master/pkg/collector/collector.go#L20

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+
+			p := &acme.Provider{
+				Configuration:  resolver.ACME,
+				Store:          localStores[resolver.ACME.Storage],
+				ChallengeStore: challengeStore,
+				ResolverName:   name,
+			}
+
+			if err := providerAggregator.AddProvider(p); err != nil {
+				log.WithoutContext().Errorf("The ACME resolver %q is skipped from the resolvers list because: %v", name, err)
```
Ullaakut, line 265:
```suggestion
				log.WithoutContext().Errorf("Skipping ACME resolver %q: %v", name, err)
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+}
+
+func setupAccessLog(conf *types.AccessLog) *accesslog.Handler {
+	if conf == nil {
+		return nil
+	}
+
+	accessLoggerMiddleware, err := accesslog.NewHandler(conf)
+	if err != nil {
+		log.WithoutContext().Warnf("Unable to create access logger : %v", err)
```
Ullaakut, line 334:
```suggestion
		log.WithoutContext().Warnf("Unable to create access logge: %v", err)
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+	// given and debug mode is disabled, the default is ERROR, and DEBUG
+	// otherwise.
+	levelStr := "error"
+	if staticConfiguration.Log != nil && staticConfiguration.Log.Level != "" {
+		levelStr = strings.ToLower(staticConfiguration.Log.Level)
+	}
+
+	level, err := logrus.ParseLevel(levelStr)
+	if err != nil {
+		log.WithoutContext().Errorf("Error getting level: %v", err)
```
Ullaakut, line 356:
```suggestion
		log.WithoutContext().Errorf("Error getting log level: %v", err)
```

cmd/proxy/proxy.go
```diff
@@ -0,0 +1,426 @@
...
+
+		if err = os.MkdirAll(dir, 0755); err != nil {
+			log.WithoutContext().Errorf("Failed to create log path %s: %s", dir, err)
+		}
+
+		err = log.OpenFile(logFile)
+
+		logrus.RegisterExitHandler(func() {
+			if closeErr := log.CloseFile(); closeErr != nil {
+				log.WithoutContext().Errorf("Error while closing log: %v", closeErr)
```
Ullaakut, line 388:
```suggestion
				log.WithoutContext().Errorf("Error while closing log file: %v", closeErr)
```

kevinpollet:
👋 Hi folks 

Good news, I've found why the `SMISuite.TestSMIAccessControlPrepareFail` test is failing (not easy though). This test expects the Maesh `prepare` cmd to fail because of an [invalid configuration in the cluster](https://github.com/dtomcej/maesh/blob/issue-492/integration/resources/smi/access-control-broken/1.tts.yaml#L9). The `TrafficTarget` destination name should be a `string`, as shown in a passing test output:

```
[]v1alpha1.TrafficTarget: v1alpha1.TrafficTarget.Destination: v1alpha1.IdentityBindingSubject.Name: 
fuzzyStringDecoder: not number or string or bool, error found in #10 byte of ...|t","name":true,"name|..., bigger context ...|a1","destination":
{"kind":"ServiceAccount","name":true,"namespace":"test"},"kind":"TrafficTarget","m|
```

This PR uses Traefik as a library instead of the Docker image for Maesh data planes. Traefik dependencies are now in the Maesh compilation path because of the [proxy](https://github.com/dtomcej/maesh/blob/issue-492/cmd/proxy/proxy.go) command. Traefik depends on the ACME client [lego](https://github.com/go-acme/lego) (used by the [static config](https://github.com/containous/traefik/blob/master/pkg/config/static/static_config.go#L11)) which depends on => [aliyun/alibaba-cloud-sdk-go](https://github.com/aliyun/alibaba-cloud-sdk-go) which depends on => [github.com/json-iterator/go](github.com/json-iterator/go) (also used by one of the [servicemeshinterface/smi-sdk-go](https://github.com/servicemeshinterface/smi-sdk-go) dependency).

The problem is that the `aliyun/alibaba-cloud-sdk-go` dependency [overrides the default string decoder of json-iterator/go](https://github.com/aliyun/alibaba-cloud-sdk-go/blob/master/sdk/responses/json_parser.go#L31). So, the test is not working anymore because a `bool` is now unmarshalled as a `string`.

To fix this test we can:

- Change the `TrafficTarget` config to make prepare cmd fail again (tried without success). 
- Fork the `aliyun/alibaba-cloud-sdk-go` project and remove the decoder override.
- Remove the decoder override in an `init` method in the `proxy` cmd. Something like the following snippet:

```
func init()  {
	jsoniter.RegisterTypeDecoder("string", nil)
}
```

BTW, I think we should a look how the `json-iterator/go` should be used by the `aliyun/alibaba-cloud-sdk-go` to fix that and open a PR.

Hope I'm clear :)

kevinpollet:
I've created the issue in the `alibaba-cloud-sdk-go` repository: https://github.com/aliyun/alibaba-cloud-sdk-go/issues/412

kevinpollet:
I've created the issue in the `alibaba-cloud-sdk-go` repository: https://github.com/aliyun/alibaba-cloud-sdk-go/issues/412

dtomcej:
Awesome catch @kevinpollet, you are a wizard for finding that one.

dtomcej:
The proxy.go file was copied almost directly from the Traefik repo, so that is why it is so rough.

Definitely will work on cleaning it here.

cmd/proxy/proxy.go
jspdown: I'm sure you meant: :wink: 
```suggestion
		log.WithoutContext().Warnf("Unable to create access logger: %v", err)
```

go.mod
```diff
@@ -22,6 +24,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.1.0+incompatible
+	github.com/aliyun/alibaba-cloud-sdk-go => github.com/aliyun/alibaba-cloud-sdk-go v1.61.112
```
ldez, line 27: The `replace` is not made to force a version.

you just have to set the version in the `require` section.

kevinpollet: You're right! Sorry for that 😑

kevinpollet: After thinking a bit more about that, I've remembered that I've read the following thing in the [go module documentation](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive).

> `replace` allows the top-level module control over the exact version used for a dependency, such as:
>
> - replace example.com/some/dependency => example.com/some/dependency v1.2.3

WDYT?

cmd/proxy/proxy.go
kevinpollet: Agree with you! We should consider having our endpoint/stats for Maesh. I think it's not possible to update the [collectorURL](https://github.com/containous/traefik/blob/master/pkg/collector/collector.go#L20) with the [compilation option `-X`](https://github.com/golang/go/wiki/GcToolchainTricks#including-build-information-in-the-executable) because it is a `const`.  So, this endpoint should be configurable in Traefik, or we should implement our mechanism. For now, I will remove the stat collection and let's address that in another issue.

kevinpollet: I've created https://github.com/containous/maesh/issues/503

helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
kevinpollet: After a discussion with @SantoDE, we think that it's better to bump a major version of the Helm Chart (`v2.0.0`) and replace the `maesh.controller.image` and `maesh.mesh.image` properties by the `maesh.image` property. Even if it's possible to deprecate them the meaning of the `maesh.mesh.image` is not the same because we're not using the Traefik image anymore.

jspdown:
The `install.md` should be updated to reflect the changes made in the `value.yaml` regarding the controller image. 

```
helm install maesh helm/chart/maesh --set controller.image.pullPolicy=IfNotPresent --set controller.image.tag=latest
```

Also, I'm wondering if we shouldn't keep the `controller.image`, add the new `image` and compute the image name, tag and pullPolicy based on both values. Since this PR will probably land in the v1.2.0, we might want to keep it backward compatible. 

@kevinpollet  WDYT?

kevinpollet:
> Also, I'm wondering if we shouldn't keep the `controller.image`, add the new `image` and compute the image name, tag and pullPolicy based on both values. Since this PR will probably land in the v1.2.0, we might want to keep it backward compatible.
> 
> @kevinpollet WDYT?

As said in a previous comment, we will release a major version of the Helm Chart. So, it doesn't need to be backward compatible.

Issue #514, opened 2020-04-21
Retrieve port mapping from existing shadow services


jspdown:
#### Proposal

Stop relying on the `tcp-state-table` ConfigMap for persisting TCP port mappings and use shadow services metadata instead. 

#### Background

When a TCP service gets created, a shadow service will be created using a port mapping retrieved from the `TCPPortMapping` object. This object loads is state on startup from the `tcp-state-table` ConfigMap and each time an update happen on the shadow service, the new state will be written in the ConfigMap.

Since the shadow-service is deeply linked to the port mapping, it contains the exact same information, we are persisting the same information at 2 different locations (shadow-service name and the config-map). If one of them is not updated, the service won't be available to the user. We actually faced the issue very recently #487 and #442 .

Benefits:
- Single source of truth.
- Same technique can be used for UDP without having to manage a 2nd ConfigMap.
- Remove ConfigMap from the RBAC definition.
- Faster and safer port mapping allocation.
- #487 is not a problem anymore.

#### Workarounds

Keep doing what we do and try to keep these states as close as possible.

dtomcej:
Accidentally closed by bot.

Reopening :)

PR #523, opened 2020-04-27
Implement UDP traffic type

### LINKED ISSUE #478: UDP Support
#### Proposal

Traefik now supports UDP.

Maesh should be able to provide UDP support for services.

#### Background

Kubernetes allows for UDP services, so should maesh.

kevinpollet:
### What does this PR do?

Fixes #478.

This PR implements the `udp` traffic type, allowing the use of `UDP` services.  

For now, the `udp` traffic type does not work when ACL mode is enabled. In ACL mode, all traffic is forbidden unless it is explicitly allowed with a [TrafficTarget](https://github.com/servicemeshinterface/smi-spec/blob/master/traffic-access-control.md) and unfortunately, the SMI specification does not yet define a [Traffic Spec](https://github.com/servicemeshinterface/smi-spec/blob/master/traffic-specs.md) for `UDP`.

A [PR](https://github.com/servicemeshinterface/smi-spec/pull/151) has been opened by @SantoDE to add the `UDPRoute` to the specification and I'll create an issue to support the `UDP` `Traffic Spec` when the PR is merged.

### How to test it

1. Build and deploy Maesh. 
2. Deploy the attached [Kubernetes resources](https://github.com/containous/maesh/files/4539567/udp.zip).
3. Run the following commands:

```
$ kubectl exec -it deployment/bastion -- nc -u whoamiudp-v1.default.maesh 8080
WHO
Hostname: whoamiudp-v1-66b7d4b9b4-ttvfc
IP: 127.0.0.1
IP: ::1
IP: 10.244.2.9
IP: fe80::a878:8ff:fe6d:15a0

$ kubectl exec -it deployment/bastion -- nc -u whoamiudp-v2.default.maesh 8080
WHO
Hostname: whoamiudp-v2-6b896f88f7-rb28r
IP: 127.0.0.1
IP: ::1
IP: 10.244.1.12
IP: fe80::60fd:4dff:fed2:ea48

$ kubectl exec -it deployment/bastion -- nc -u whoamiudp.default.maesh 8080
WHO
Hostname: whoamiudp-v1-66b7d4b9b4-ttvfc
IP: 127.0.0.1
IP: ::1
IP: 10.244.2.9
IP: fe80::a878:8ff:fe6d:15a0
WHO
Hostname: whoamiudp-v2-6b896f88f7-rb28r
IP: 127.0.0.1
IP: ::1
IP: 10.244.1.12
IP: fe80::60fd:4dff:fed2:ea48
```


dtomcej:
since the two state tables will never overlap (aka they have a different starting offset), do you think that they could be combined?

Instead of having two, just use the configured ranges to determine which is which?

kevinpollet:
> since the two state tables will never overlap (aka they have a different starting offset), do you think that they could be combined?

From my point of view, using the configured range does not work because the port of `k8s.ServiceWithPort` cannot tell us if the mapped port should be in the `UDP` or `TCP` range. 

As you said, I've also thought that we could add the `type` to `k8s.ServiceWithPort` but it seems a bit weird for me to have a `PortMapper` dealing with `UDP` and `TCP` range at the same time. If we want to go for this solution, I think the `PortMapper` should also deal with the `HTTP` port mapping for consistency.

WDYT?

dtomcej:
I think that this may be less of an issue once we implement #514, as we will not need to store/retrieve this info from the configmaps anymore, as that information will come from the shadow service itself (via annotations).

I think that we can continue with this PR as-is, and review the PortMapper when we are dealing with #514

pkg/k8s/constants.go
```diff
@@ -43,11 +45,10 @@ const (
 	// SchemeHTTPS HTTPS scheme.
 	SchemeHTTPS string = "https"
 
-	// BlockAllMiddlewareKey block all middleware name.
```
dtomcej, line 46: Is this not used at all?

integration/integration_test.go
```diff
@@ -63,6 +63,7 @@ func Test(t *testing.T) {
 	images = append(images, image{"containous/maesh:latest", false})
 	images = append(images, image{"containous/whoami:v1.0.1", true})
 	images = append(images, image{"containous/whoamitcp", true})
+	images = append(images, image{"containous/whoamiudp:dev", true})
```
dtomcej, line 66: is `dev` the right tag to be using here?

pkg/provider/fixtures/annotations-traffic-type-config.json
```diff
@@ -60,5 +60,29 @@
         }
       }
     }
+  },
+  "udp": {
+    "routers": {
+      "my-ns-svc-b-8080": {
+        "entryPoints": [
+          "udp-10000"
```
dtomcej, line 68: shouldn't this be 15000?

pkg/provider/fixtures/annotations-traffic-type-topology.json
```diff
@@ -20,6 +20,27 @@
         "pod-a1@my-ns",
         "pod-a2@my-ns"
       ]
+    },
+    "svc-b@my-ns": {
+      "name": "svc-b",
+      "namespace": "my-ns",
+      "selector": {},
+      "annotations": {
+        "maesh.containo.us/traffic-type": "udp"
```
dtomcej, line 29: Is this annotation required? since the port is specifically UDP?

pkg/provider/provider.go
```diff
@@ -179,6 +181,20 @@ func (p *Provider) buildServicesAndRoutersForService(t *topology.Topology, cfg *
 			cfg.TCP.Services[key] = p.buildTCPServiceFromService(t, svc, svcPort.TargetPort.IntVal)
 			cfg.TCP.Routers[key] = buildTCPRouter(rule, entrypoint, key)
 		}
+
+	case k8s.ServiceTypeUDP:
```
dtomcej, line 185: What happens if a service has both TCP and UDP ports?

Or the default type is TCP, and a user adds a UDP port?

pkg/k8s/constants.go
kevinpollet: Yes

integration/integration_test.go
kevinpollet: The `latest` tag does not exist for the [containous/whoamiudp](https://hub.docker.com/r/containous/whoamiudp/tags) image.

pkg/provider/fixtures/annotations-traffic-type-config.json
kevinpollet: Yes, to be consistent.

pkg/provider/fixtures/annotations-traffic-type-topology.json
kevinpollet: If the `maesh.containo.us/traffic-type` annotation is not defined, Maesh uses the configured default traffic type mode. So, if the default mode is `HTTP` the `UDP` ports will not be available through Maesh.

pkg/provider/provider.go
kevinpollet: To be consistent with the traffic type definition, if the service `traffic-type` is `http` or `tcp` the `UDP` ports are not available through Maesh. If the service traffic type is `udp` only the UDP ports will be available.

In my humble opinion, this behaviour seems a bit weird because sometimes it doesn't make sense for a service to have only one traffic type (eg. a DNS service uses TCP and UDP). This limitation already exists if your service uses TCP for a custom app protocol and HTTP (you'll have to set the `traffic-type` to `tcp` and cannot apply HTTP middlewares).

For now, the user has to create two services to bypass this limitation but this is not acceptable from my point of view. Kubernetes allows a service to have ports with a different protocol so we should support that out of the box.

Maybe the `traffic-type` annotation should only be used to specify the traffic type for a given port to apply specific middlewares and so on (I'll create the issue if you're ok with that).

For UDP, maybe I should have made UDP port available by default but this behaviour seems not consistent to me with the current `traffic-type` behaviour.

WDYT?

cmd/maesh/maesh.go
```diff
@@ -104,6 +105,8 @@ func maeshCommand(iConfig *cmd.MaeshConfiguration) error {
 		APIHost:          iConfig.APIHost,
 		MinTCPPort:       minTCPPort,
 		MaxTCPPort:       minTCPPort + iConfig.LimitTCPPort,
+		MinUDPPort:       minUDPPort,
```
Ullaakut, line 108: I am slightly bothered that we don't always list TCP, UDP and HTTP in the same order :p

To be consistent with the dashboard (maybe?) WDYT of always going HTTP>TCP>UDP?

docs/content/configuration.md
```diff
@@ -45,9 +45,14 @@ The traffic type can be configured by using the following annotation:
 maesh.containo.us/traffic-type: "http"
 ```
 
-This annotation can be set to either `http` or `tcp`, and will specify the mode for that service operation.
+This annotation can be set to either `http`, `tcp` or `udp` and will specify the mode for that service operation.
```
Ullaakut, line 48:
```suggestion
This annotation can be set to either `http`, `tcp` or `udp` and specifies the mode for that service operation.
```

integration/integration_test.go
Ullaakut: Would be neat to push one

pkg/controller/controller.go
```diff
@@ -128,7 +137,7 @@ func (c *Controller) init() {
 	c.splitFactory = splitInformer.NewSharedInformerFactoryWithOptions(c.clients.GetSplitClient(), k8s.ResyncPeriod)
 
 	c.ServiceLister = c.kubernetesFactory.Core().V1().Services().Lister()
-	c.serviceManager = NewShadowServiceManager(c.logger, c.ServiceLister, c.cfg.Namespace, c.tcpStateTable, c.cfg.DefaultMode, c.cfg.MinHTTPPort, c.cfg.MaxHTTPPort, c.clients.GetKubernetesClient())
+	c.serviceManager = NewShadowServiceManager(c.logger, c.ServiceLister, c.cfg.Namespace, c.tcpStateTable, c.udpStateTable, c.cfg.DefaultMode, c.cfg.MinHTTPPort, c.cfg.MaxHTTPPort, c.clients.GetKubernetesClient())
```
Ullaakut, line 140: Not in this PR but it could be nice to improve dependency management for the shadow service mgr. This is painful, hehe

pkg/k8s/constants.go
jspdown: Good catch :eagle:

cmd/maesh/maesh.go
kevinpollet: Sure :)

pkg/provider/provider.go
jspdown: I'm also feeling bad about this limitation. We are actually facing the same situation with middlewares. It would be great to all user to specify `traffic-type` as well as middlewares for a specific port. I'm thinking about something like this: 
```
// Applies on every ports
maesh.containo.us/traffic-type: "tcp"
// Applies only on the port named "api"
maesh.containo.us/api/traffic-type: "http"
maesh.containo.us/api/circuit-breaker-expression: "..."
```

pkg/controller/controller.go
jspdown: It might be interesting to promote the `Config` struct from the provider package. At the end it represents Maesh controller configuration.

pkg/provider/provider.go
kevinpollet: I'm also thinking about something like that. I'll try to write a proposal on that subject.

pkg/controller/service.go
```diff
@@ -160,11 +169,12 @@ func (s *ShadowServiceManager) cleanupPortMapping(oldUserSvc *corev1.Service, ne
 		}
```
jspdown, line 169: Shouldn't we do something specific for UDP here? If the port.Protocol isn't UDP, we shouldn't consider it found. no?

pkg/provider/provider.go
```diff
@@ -53,18 +53,20 @@ type Provider struct {
...
+	tcpStateTable          PortFinder
+	udpStateTable          PortFinder
 	buildServiceMiddleware MiddlewareBuilder
 
 	logger logrus.FieldLogger
 }
 
 // New creates a new Provider.
-func New(topologyBuilder TopologyBuilder, tcpStateTable TCPPortFinder, cfg Config, logger logrus.FieldLogger) *Provider {
+func New(topologyBuilder TopologyBuilder, tcpStateTable PortFinder, udpStateTable PortFinder, cfg Config, logger logrus.FieldLogger) *Provider {
```
jspdown, line 64:
```suggestion
func New(topologyBuilder TopologyBuilder, tcpStateTable, udpStateTable PortFinder, cfg Config, logger logrus.FieldLogger) *Provider {
```

integration/integration_test.go
kevinpollet: After thinking a bit more about that, I've used the `digest` tag for the `whoamitcp` and `whoamiudp` images to make sure that the build is reproducible. We'll have to update the image tags when a version tags will be available.

pkg/provider/provider.go
```diff
@@ -179,6 +181,20 @@ func (p *Provider) buildServicesAndRoutersForService(t *topology.Topology, cfg *
 			cfg.TCP.Services[key] = p.buildTCPServiceFromService(t, svc, svcPort.TargetPort.IntVal)
 			cfg.TCP.Routers[key] = buildTCPRouter(rule, entrypoint, key)
 		}
+
+	case k8s.ServiceTypeUDP:
+		for _, svcPort := range svc.Ports {
+			entrypoint, err := p.buildUDPEntrypoint(svc, svcPort.Port)
+			if err != nil {
+				p.logger.Errorf("Unable to build UDP entrypoint for Service %q and port %d: %v", svcKey, svcPort.Port, err)
```
jspdown, line 189: If I understand this correctly. When creating the shadow service, UDP ports have been saved into the UDP state table. And so, while iterating on ports here, if the port is not UDP, we are just not gonna find it in the UDP state table. That's how ports get filtered. Could you confirm?

pkg/provider/provider.go
```diff
@@ -272,82 +288,119 @@ func (p *Provider) buildServiceAndRoutersForTrafficSplit(t *topology.Topology, c
...
-					Name:   backendSvcKey,
-					Weight: getIntRef(backend.Weight),
-				}
-			}
 
-			entrypoint, err := p.buildTCPEntrypoint(tsSvc, svcPort.Port)
-			if err != nil {
-				p.logger.Errorf("Unable to build TCP entrypoint for TrafficTarget %q and port %d: %v", tsKey, svcPort.Port, err)
-				continue
+		entrypoint, err := p.buildTCPEntrypoint(tsSvc, svcPort.Port)
```
jspdown, line 369: :thinking: I think there is a problem here. If the port is `UDP`, we are going to build backend services before we notice its not a TCP port. So we will end up with orphans backend services.
In order to fix that kind of bug, we must call `buildTCPEntrypoint` before doing anything. Another option would be to be more explicit on which  protocol we are interested in. This could be done by checking the `svcPort.Protocol != "TCP"`  before iterating on `ts.Backends`. This kind of issue is not obvious to detect. I suggest we filter ports in these loops everywhere.

pkg/provider/provider.go
```diff
@@ -272,82 +288,119 @@ func (p *Provider) buildServiceAndRoutersForTrafficSplit(t *topology.Topology, c
...
+			backendSvcs[i] = dynamic.UDPWRRService{
+				Name:   backendSvcKey,
+				Weight: getIntRef(backend.Weight),
 			}
+		}
 
-			key := getServiceRouterKeyFromService(tsSvc, svcPort.Port)
-			cfg.TCP.Services[key] = buildTCPServiceFromTrafficSplit(backendSvcs)
-			cfg.TCP.Routers[key] = buildTCPRouter(tcpRule, entrypoint, key)
+		entrypoint, err := p.buildUDPEntrypoint(tsSvc, svcPort.Port)
```
jspdown, line 394: Same issue here, backend services will be created even for TCP ports.

pkg/controller/service.go
kevinpollet: Good catch!

pkg/provider/provider.go
kevinpollet: Yes.

pkg/provider/provider.go
kevinpollet: See above.

pkg/provider/provider.go
kevinpollet: You're right! I think we have to be explicit because we're creating false errors in my point of view and that should be addressed in another PR/Issue. For now, as you said I've called `buildTCPEntrypoint` and `buildUDPEntrypoint` before doing anything.

PR #525, opened 2020-04-27
Add HTTP provider to proxy nodes

### LINKED ISSUE #493: Add HTTP Provider to Proxy nodes
#### Proposal

We should add an HTTP polling provider to the proxy nodes so that we don't have to deploy code to nodes, instead they can watch for configuration updates.

#### Background

The deploy code is not optimal, and does not scale or perform as well as we had hoped.

dtomcej:
## This PR:

- Adds HTTP Provider
- Adds k8s service for controller API endpoint
- Configures maesh proxy nodes to use the API endpoint for HTTP provider
- Removes deploy code from controller
- Removes deploylog package

Fixes #493 

## Additional Notes:

This paves the way to allowing each proxy node to get its own configuration. This can be done right now using the named endpoints, but until there are actually different configurations built, it doesn't make sense to use those endpoints at this time.


pkg/config/static/static_config.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/containous/traefik/v2/pkg/provider/rest"
+	"github.com/containous/traefik/v2/pkg/types"
+)
+
+// DefaultInternalEntryPointName the name of the default internal entry point
+const DefaultInternalEntryPointName = "traefik"
+
+// Configuration is the static configuration
+type Configuration struct {
+	Global *traefikStatic.Global `description:"Global configuration options" json:"global,omitempty" toml:"global,omitempty" yaml:"global,omitempty" export:"true"`
```
Ullaakut, line 17: Why not embed a `traefikStatic.Configuration` instead of duplicating all of its fields here? Are we planning on deleting some or just adding new ones?

cmd/proxy/proxy.go
```diff
@@ -124,39 +128,49 @@ func runCmd(staticConfiguration *static.Configuration) error {
 	return nil
 }
 
-func setupServer(staticConfiguration *static.Configuration) (*server.Server, error) {
-	providerAggregator := aggregator.NewProviderAggregator(*staticConfiguration.Providers)
+func setupServer(maeshStaticConfiguration *static.Configuration) (*server.Server, error) {
+	traefikStaticConfiguration := maeshStaticConfiguration.ToTraefikConfig()
```
Ullaakut, line 132: The `ToTraefikConfig()` and having two variables could be solved by embedding the traefik configuration in the proxy configuration directly. You'd then use `proxyCfg.Configuration` to access the traefik config, wdyt?

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+func (p *Provider) getDataFromEndpoint(ctx context.Context) ([]byte, error) {
+	b := []byte{}
+	req, err := http.NewRequest(http.MethodGet, p.Endpoint, bytes.NewBuffer(b))
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to create request: %w", err)
+	}
+
+	client := &http.Client{Timeout: time.Duration(p.PollTimeout)}
+	resp, err := client.Do(req)
```
jspdown, line 124: WDYT of just using `client.Get` and getting rid of the `b`, `req` and it's error handling?

pkg/config/static/static_config.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	httpProvider "github.com/containous/maesh/pkg/proxy/provider/http"
+	traefikStatic "github.com/containous/traefik/v2/pkg/config/static"
+	"github.com/containous/traefik/v2/pkg/ping"
+	"github.com/containous/traefik/v2/pkg/provider/file"
+	"github.com/containous/traefik/v2/pkg/provider/rest"
+	"github.com/containous/traefik/v2/pkg/types"
+)
+
+// DefaultInternalEntryPointName the name of the default internal entry point
```
Ullaakut, line 12:
```suggestion
// DefaultInternalEntryPointName is the name of the default internal entry point
```

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+		if resp.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("received non-ok response code: %d", resp.StatusCode)
+		}
+
+		log.FromContext(ctx).Debugf("Successfully received data from endpoint: %q", p.Endpoint)
+
+		return bodyData, nil
+	}
+
+	if err != nil {
```
jspdown, line 147: WDYT of handling errors first? 

```
resp, err := client.Do(req)
if err != nil {
    return nil, fmt.Errorf("unable to get data from endpoint: %w", err)
}
if resp == nil {
    return nil, fmt.Errorf("unable to get data from endpoint: %w", err)
} 

// handle success case here
```

I think it would make the code easier to read.

pkg/config/static/static_config.go
```diff
@@ -0,0 +1,90 @@
...
+const DefaultInternalEntryPointName = "traefik"
+
+// Configuration is the static configuration
```
Ullaakut, line 15:
```suggestion
// Configuration is the Maesh proxy's static configuration
```

Could make sense to try to differentiate it to the traefik static configuration, wdyt?

pkg/config/static/static_config.go
```diff
@@ -0,0 +1,90 @@
...
+		if _, ok := c.EntryPoints[DefaultInternalEntryPointName]; !ok {
+			ep := &traefikStatic.EntryPoint{Address: ":8080"}
+			ep.SetDefaults()
+			c.EntryPoints[DefaultInternalEntryPointName] = ep
+		}
+	}
+}
+
+// ToTraefikConfig returns a Traefik compatable configuration for use in the proxy code without affecting compatibility.
+func (c *Configuration) ToTraefikConfig() *traefikStatic.Configuration {
```
Ullaakut, line 70: Can be removed if the traefik config is embedded in the `Configuration` type

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return nil
+}
+
+// Provide allows the provider to provide configurations to traefik
+// using the given configuration channel.
+func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.Pool) error {
+	pool.GoCtx(func(routineCtx context.Context) {
+		ctxLog := log.With(routineCtx, log.Str(log.ProviderName, "http"))
```
Ullaakut, line 60: WDYT of having a const called `ProviderName` that contains http? We're passing it in a few places here, it could be cleaner.

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+		err := backoff.RetryNotify(safe.OperationWithRecover(operation), backoff.WithContext(job.NewBackOff(backoff.NewExponentialBackOff()), ctxLog), notify)
+		if err != nil {
+			logger.Errorf("Cannot connect to http server %w", err)
+		}
+	})
+
+	return nil
+}
+
+// getDataFromEndpoint gets data from the configured provider endpoint, and returns the data.
```
Ullaakut, line 114:
```suggestion
// getDataFromEndpoint returns data from the configured provider endpoint.
```

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+func (p *Provider) getDataFromEndpoint(ctx context.Context) ([]byte, error) {
+	b := []byte{}
+	req, err := http.NewRequest(http.MethodGet, p.Endpoint, bytes.NewBuffer(b))
+
```
Ullaakut, line 118:
```suggestion
```

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+	if resp != nil {
+		defer resp.Body.Close()
+
+		var (
+			bodyData []byte
+			bodyErr  error
+		)
+
+		if bodyData, bodyErr = ioutil.ReadAll(resp.Body); bodyErr != nil {
```
jspdown, line 134: `err` could be reused if the error case of the `client.Do` is handled first.

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+	})
+
+	return nil
+}
+
+// getDataFromEndpoint gets data from the configured provider endpoint, and returns the data.
+func (p *Provider) getDataFromEndpoint(ctx context.Context) ([]byte, error) {
+	b := []byte{}
+	req, err := http.NewRequest(http.MethodGet, p.Endpoint, bytes.NewBuffer(b))
```
Ullaakut, line 117: You can create a bytes.Buffer directly without having to provide an empty byte slice

```suggestion
	req, err := http.NewRequest(http.MethodGet, p.Endpoint, &bytes.Buffer{})
```

If you need to get its contents later on, define it as a variable and use `buffer.Bytes()` or `buffer.String()`

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+			logger.Errorf("Cannot connect to http server %w", err)
+		}
+	})
+
+	return nil
+}
+
+// getDataFromEndpoint gets data from the configured provider endpoint, and returns the data.
+func (p *Provider) getDataFromEndpoint(ctx context.Context) ([]byte, error) {
+	b := []byte{}
```
Ullaakut, line 116:
```suggestion
```

Can be removed IMO, see next suggestion

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+	req, err := http.NewRequest(http.MethodGet, p.Endpoint, bytes.NewBuffer(b))
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to create request: %w", err)
+	}
+
+	client := &http.Client{Timeout: time.Duration(p.PollTimeout)}
+	resp, err := client.Do(req)
+
+	if resp != nil {
```
Ullaakut, line 126: WDYT of checking `err` before `resp`?

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+			return nil, fmt.Errorf("received non-ok response code: %d", resp.StatusCode)
+		}
+
+		log.FromContext(ctx).Debugf("Successfully received data from endpoint: %q", p.Endpoint)
+
+		return bodyData, nil
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to get data from endpoint: %w", err)
```
Ullaakut, line 148: WDYT of adding the actual endpoint in the error, to make debugging easier?

```suggestion
		return nil, fmt.Errorf("unable to get data from endpoint %q: %w", p.Endpoint, err)
```

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return nil, fmt.Errorf("received no data from endpoint")
```
Ullaakut, line 151: Instead of checking if `resp != nil` above and if not to return an error at the end of the func, what do you think of failing fast and writing

```go
		if resp == nil {
			return nil, fmt.Errorf("received no data from endpoint")
		}

		defer resp.Body.Close()

		var (
			bodyData []byte
			bodyErr  error
		)

		if bodyData, bodyErr = ioutil.ReadAll(resp.Body); bodyErr != nil {
			return nil, fmt.Errorf("unable to read response body: %w", bodyErr)
		}

		if resp.StatusCode != http.StatusOK {
			return nil, fmt.Errorf("received non-ok response code: %d", resp.StatusCode)
		}

		log.FromContext(ctx).Debugf("Successfully received data from endpoint: %q", p.Endpoint)

		return bodyData, nil
```

instead? This way the natural flow of the function is preserved, and the expected valid outcome is at the end of the function, instead of nested within an `if` statement

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// buildConfiguration builds a configuration from the provided data.
+func (p *Provider) buildConfiguration(ctx context.Context, data []byte) *dynamic.Configuration {
```
Ullaakut, line 155: WDYT of calling it `unmarshalConfiguration` then? `build` suggests that there is some logic to this function, while it just calls `json.Unmarshal`. Also, is it really necessary to have a function for that? is it a method that we use from many places in the code?

Finally, does it need to be a method on the provider struct? It seems to me like it doesn't need to be attached to anything, as it is.

```suggestion
func unmarshalConfiguration(ctx context.Context, data []byte) *dynamic.Configuration {
```

pkg/proxy/provider/http/http_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"testing"
+	"time"
+
+	"github.com/containous/traefik/v2/pkg/config/dynamic"
+	"github.com/containous/traefik/v2/pkg/safe"
+	"github.com/containous/traefik/v2/pkg/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProviderInit(t *testing.T) {
```
Ullaakut, line 16:
```suggestion
func TestProvider_Init(t *testing.T) {
```

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+		if err != nil {
+			logger.Errorf("Cannot connect to http server %w", err)
+		}
+	})
+
+	return nil
+}
+
+// getDataFromEndpoint gets data from the configured provider endpoint, and returns the data.
+func (p *Provider) getDataFromEndpoint(ctx context.Context) ([]byte, error) {
```
Ullaakut, line 115:
```suggestion
func (p Provider) getDataFromEndpoint(ctx context.Context) ([]byte, error) {
```

The pointer receiver does not seem necessary

pkg/proxy/provider/http/http_test.go
```diff
@@ -0,0 +1,88 @@
...
+	provider := &Provider{}
+	assert.Error(t, provider.Init())
+
+	provider = &Provider{
+		Endpoint: "localhost",
+	}
+	assert.NoError(t, provider.Init())
+}
+
+func TestGetDataFromEndpoint(t *testing.T) {
```
Ullaakut, line 26:
```suggestion
func TestProvider_GetDataFromEndpoint(t *testing.T) {
```

pkg/proxy/provider/http/http_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	assert.NoError(t, provider.Init())
+
+	data, err := provider.getDataFromEndpoint(context.Background())
+	assert.NoError(t, err)
+	assert.Equal(t, "{OK}", string(data))
+}
+
+func TestBuildConfiguration(t *testing.T) {
```
Ullaakut, line 46:
```suggestion
func TestProvider_BuildConfiguration(t *testing.T) {
```

pkg/proxy/provider/http/http_test.go
```diff
@@ -0,0 +1,88 @@
...
+	provider := Provider{
+		Endpoint: "http://127.0.0.1:9000/endpoint",
+	}
+
+	assert.NoError(t, provider.Init())
+
+	config := provider.buildConfiguration(context.Background(), []byte("{}"))
```
Ullaakut, line 53: Is this test really relevant? 😛 

Maybe we don't need to test this func since it only calls `json.Unmarshal` at the moment, wdyt?

pkg/proxy/provider/http/http_test.go
```diff
@@ -0,0 +1,88 @@
...
+	assert.NotEqual(t, nil, config)
+}
+
+func TestProvide(t *testing.T) {
```
Ullaakut, line 57:
```suggestion
func TestProvider_Provide(t *testing.T) {
```

pkg/proxy/provider/http/http_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	go func() {
+		err := provider.Provide(configChan, safe.NewPool(context.Background()))
+		assert.NoError(t, err)
+	}()
+
+	timeout := time.After(time.Second)
+	select {
+	case conf := <-configChan:
+		assert.NotNil(t, conf.Configuration)
```
Ullaakut, line 84: Don't we want to check that with an endpoint that returns an actual config, we provide the expected config?

helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
```diff
@@ -73,8 +73,9 @@ spec:
             - {{ printf "\"--entryPoints.tcp-%d.address=:%d\"" $port $port }}
             - {{ printf "\"--entryPoints.tcp-%d.forwardedHeaders.insecure=true\"" $port }}
             {{- end }}
-            - "--providers.rest"
-            - "--providers.rest.insecure"
+            - "--providers.http"
+            - {{ printf "--providers.http.endpoint=http://maesh-controller.%s.svc.%s:9000/api/configuration/current" .Release.Namespace (default "cluster.local" .Values.clusterDomain) | quote }}
+            - "--providers.http.pollInterval=100ms"
```
jspdown, line 78: Could we make this configurable in the `value.yaml`?

cmd/proxy/proxy.go
```diff
@@ -35,53 +37,55 @@ import (
...
 	}
 
 	log.WithoutContext().Infof("Maesh proxy version %s built on %s", version.Version, version.Date)
 
-	jsonConf, err := json.Marshal(staticConfiguration)
+	jsonConf, err := json.Marshal(traefikStaticConfiguration)
 	if err != nil {
 		log.WithoutContext().Errorf("Could not marshal static configuration: %v", err)
-		log.WithoutContext().Debugf("Static configuration loaded [struct] %#v", staticConfiguration)
+		log.WithoutContext().Debugf("Static configuration loaded [struct] %#v", traefikStaticConfiguration)
```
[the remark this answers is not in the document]
kevinpollet: To be sure that I'm not missing something.
Why do we have to marshal the static conf here?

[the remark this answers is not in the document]
Ullaakut: Just to display it in the logs :p

cmd/proxy/proxy.go
```diff
@@ -35,53 +37,55 @@ import (
...
 	}
 
-	svr, err := setupServer(staticConfiguration)
+	svr, err := setupServer(proxyStaticConfiguration)
 	if err != nil {
 		return err
 	}
 
-	ctx := cmd.ContextWithSignal(context.Background())
+	ctx := traefikCmd.ContextWithSignal(context.Background())
```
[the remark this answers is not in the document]
dtomcej: yes, its a pretty small piece of code IIRC, so we could just use our own

pkg/config/static/static_config.go
dtomcej: We are deleting a large portion of this configuration (in particular the providers)

cmd/proxy/proxy.go
```diff
@@ -35,53 +37,55 @@ import (
...
 	if err != nil {
 		log.WithoutContext().Errorf("Could not marshal static configuration: %v", err)
-		log.WithoutContext().Debugf("Static configuration loaded [struct] %#v", staticConfiguration)
+		log.WithoutContext().Debugf("Static configuration loaded [struct] %#v", traefikStaticConfiguration)
 	} else {
 		log.WithoutContext().Debugf("Static configuration loaded %s", string(jsonConf))
 	}
 
-	svr, err := setupServer(staticConfiguration)
+	svr, err := setupServer(proxyStaticConfiguration)
```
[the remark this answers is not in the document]
dtomcej: Good catch.

cmd/proxy/proxy.go
dtomcej: The providers are different

dtomcej: and all of the middlewarebuilders etc in this function take the traefik static config as arguments, so it is easier at this point to output a compatable struct

pkg/proxy/provider/http/http.go
dtomcej: Good call. Removed.

pkg/config/static/static_config.go
```diff
@@ -0,0 +1,90 @@
...
+	return &traefikStatic.Configuration{
+		Global:           c.Global,
+		ServersTransport: c.ServersTransport,
+		EntryPoints:      c.EntryPoints,
+		Providers: &traefikStatic.Providers{
+			ProvidersThrottleDuration: c.Providers.ProvidersThrottleDuration,
+			// These are the two providers that we would provide configuration for at this time.
+			File: c.Providers.File,
+			Rest: c.Providers.Rest,
```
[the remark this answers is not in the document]
dtomcej: upon review, I don't think we do.

pkg/config/static/static_config.go
dtomcej: I have rethought this, and I might be able to make it work

cmd/proxy/proxy.go
dtomcej: fixed.

helm/chart/maesh/templates/controller/controller-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+    component: controller
+spec:
+  type: ClusterIP
+  ports:
+    - port: 9000
+      name: mesh-controller-api
+      targetPort: api
+  selector:
+    component: controller
+
```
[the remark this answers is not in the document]
dtomcej: I suppose, if we have to.

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+			// channel closed
+			return nil
+		}
+
+		notify := func(err error, time time.Duration) {
+			logger.Errorf("Provider connection error %w, retrying in %s", err, time)
+		}
+		err := backoff.RetryNotify(safe.OperationWithRecover(operation), backoff.WithContext(job.NewBackOff(backoff.NewExponentialBackOff()), ctxLog), notify)
+		if err != nil {
+			logger.Errorf("Cannot connect to http server %w", err)
```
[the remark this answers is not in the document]
dtomcej: isn't `%w` the correct way to wrap an error now?

pkg/controller/controller.go
```diff
@@ -219,11 +208,7 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 			if message == k8s.ConfigMessageChanForce || !reflect.DeepEqual(c.lastConfiguration.Get(), conf) {
 				c.lastConfiguration.Set(conf)
 
-				if deployErr := c.deployConfiguration(conf); deployErr != nil {
-					break
-				}
-
-				// Configuration successfully deployed, enable readiness in the api.
+				// Configuration successfully created, enable readiness in the api.
 				c.api.EnableReadiness()
```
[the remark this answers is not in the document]
dtomcej: It just means that it has successfully built a configuration now.

pkg/controller/controller.go
```diff
@@ -232,19 +217,13 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 				break
 			}
 
-			dynCfg, ok := rawCfg.(*dynamic.Configuration)
+			_, ok := rawCfg.(*dynamic.Configuration)
```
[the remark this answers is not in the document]
dtomcej: nope.

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+
+// Provider is a provider.Provider implementation that queries an endpoint for a configuration.
+type Provider struct {
+	Endpoint     string         `description:"Load configuration from this endpoint." json:"endpoint" toml:"endpoint" yaml:"endpoint" export:"true"`
+	PollInterval types.Duration `description:"Polling interval for endpoint." json:"pollInterval,omitempty" toml:"pollInterval,omitempty" yaml:"pollInterval,omitempty"`
+	PollTimeout  types.Duration `description:"Polling timeout for endpoint." json:"pollTimeout,omitempty" toml:"pollTimeout,omitempty" yaml:"pollTimeout,omitempty"`
+}
+
+// New creates a new instance of the HTTP provider.
+func New(endpoint string, pollInterval types.Duration, pollTimeout types.Duration) *Provider {
```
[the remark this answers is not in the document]
dtomcej: This provider is used by the traefik core, and I'm not sure this would work with that logging method would it? We don't have a fieldlogger in the proxy node.

pkg/proxy/provider/http/http.go
```diff
@@ -0,0 +1,164 @@
...
+func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.Pool) error {
+	pool.GoCtx(func(routineCtx context.Context) {
+		ctxLog := log.With(routineCtx, log.Str(log.ProviderName, "http"))
+		logger := log.FromContext(ctxLog)
+
+		operation := func() error {
+			errChan := make(chan error)
+			ticker := time.NewTicker(time.Duration(p.PollInterval))
+
+			pool.GoCtx(func(ctx context.Context) {
```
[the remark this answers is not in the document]
dtomcej: I took a Traefik provider, and stripped back a lot of what I thought was unneeded, but TBH there is still a bunch here that I am not confident with how it works.

cmd/configuration.go
```diff
@@ -58,3 +65,27 @@ func NewPrepareConfig() *PrepareConfig {
...
 }
+
+// ProxyConfiguration wraps the static configuration and extra parameters for proxy nodes.
+type ProxyConfiguration struct {
+	// ConfigFile is the path to the configuration file.
+	static.Configuration `export:"true"`
+}
+
+// NewProxyConfiguration creates a ProxyConfiguration with default values.
+func NewProxyConfiguration() *ProxyConfiguration {
```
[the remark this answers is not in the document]
dtomcej: https://github.com/containous/maesh/issues/534

cmd/proxy/proxy.go
```diff
@@ -28,135 +28,108 @@ import (
...
+	log.WithoutContext().Debugf("Static configuration loaded %s", string(jsonConf))
+
+	svr, err := setupServer(proxyConfiguration)
 	if err != nil {
 		return err
 	}
 
 	ctx := cmd.ContextWithSignal(context.Background())
 
-	if staticConfiguration.Ping != nil {
```
kevinpollet, line 83: The CLI flag in the Helm chart here: https://github.com/containous/maesh/blob/master/helm/chart/maesh/templates/mesh/mesh-daemonset.yaml#L157 should be removed too.

dtomcej: This was removed by accident, It has been replaced and updated.

kevinpollet: I thought it was unused but it is by the readiness service.

pkg/proxy/provider/http/http.go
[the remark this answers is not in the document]
dtomcej: I'm wondering if we should rework this whole method, since i'm sure we could strip 90% of this code. I'm not sure if we need all of this code.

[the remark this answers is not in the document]
dtomcej: @jspdown would you be able to assist me with a follow up issue to refactor this method?

pkg/proxy/provider/http/http.go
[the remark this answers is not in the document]
jspdown: Ah! true!

pkg/proxy/provider/http/http.go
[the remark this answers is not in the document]
jspdown: Yes sure! @kevinpollet has created an issue for that.

PR #527, opened 2020-04-28
Remove dependency from provider to topology builder

Ullaakut:
### What does this PR do?

This PR removes the dependency from the provider to a topology builder, by making `BuildConfig` take a topology directly.

This means that it's now the controller's responsibility to call the topology builder and to pass that configuration to the provider.

### How to test it

* `make test-integration` should still work as expected
* `make test` should also work

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -42,6 +42,11 @@ type TCPPortMapper interface {
 	Remove(svc k8s.ServiceWithPort) (int32, error)
 }
 
+// TopologyBuilder builds Topologies.
+type TopologyBuilder interface {
+	Build(ignoredResources k8s.IgnoreWrapper) (*topology.Topology, error)
+}
+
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -210,12 +217,14 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 			return nil
 		case message := <-c.configRefreshChan:
 			// Reload the configuration.
-			conf, confErr := c.provider.BuildConfig()
-			if confErr != nil {
-				c.logger.Errorf("Unable to build dynamic configuration: %v", confErr)
+			topo, err := c.topologyBuilder.Build(c.ignored)
```
skwair, line 220: `s/ignored/ignoredSomething`

Can we rename this variable to explicitly state what we ignore, is it kubernetes resources?

PR #528, opened 2020-04-28
Remove deprecated SMI option in the Helm chart

### LINKED ISSUE #502: Remove the deprecated SMI option in Helm chart
#### Proposal

Due to #495, we're planning to bump a new major version of the helm chart (`v2.0.0`). It's maybe a good time to remove the newly deprecated `SMI Mode`.

kevinpollet:
### What does this PR do?

Fixes #502.

This PR removes the deprecated SMI option in the Helm chart.

### Additional Notes

This PR removes also the Traefik EE reference in the Helm chart guidelines.

### DIFF
--- a/docs/content/install.md
+++ b/docs/content/install.md
@@ -15,7 +15,7 @@ helm install maesh maesh/maesh
 
 ## Install from source
 
-!!! Note Supported Installations
+!!! Note "Supported Installations"
     Please be aware that the supported installation method is via Helm, using official releases.
     If you want to build/install/run Maesh from source, we may not be able to provide support.
     Installing from source is intended for development/contributing.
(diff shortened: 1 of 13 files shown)


docs/content/install.md
```diff
@@ -73,26 +73,15 @@ helm install maesh --namespace=maesh maesh/maesh --set acl=true
...
-To enable SMI, install maesh in SMI mode by setting the smi.enable helm chart option to true.
-
-```bash
-helm install maesh --namespace=maesh maesh/maesh --set smi.enable=true`
-```
-
-The smi.enable option makes Maesh process SMI resources.
-
-!!! Note This option is deprecated. Please consider using `--acl` instead.
+!!! Warning "SMI Mode (Deprecated)"
```
jspdown, line 82: Can we still call it "deprecated" if the option has been removed? If this is no longer a valid option, couldn't we remove it from the documentation. We might want to release a document to explain the migration v1->v2.

kevinpollet: That was also my feeling, but as this option has been removed without a real deprecation, I think we should keep this note for now.

@SantoDE thoughts?

dtomcej: Should we have a migration page in the docs like traefik has? https://docs.traefik.io/migration/v2/

That would allow us to make the changes we need, and document them.

SantoDE: +1 on that @dtomcej

PR #529, opened 2020-04-28
Inject errors in topology

### LINKED ISSUE #522: Inject errors in topology
Replaces #477 

In the end, we'd rather inject topology builder errors in the built topology, just like we do it for Traefik's dynamic configuration already. This makes the process of pointing out exactly where an issue lies much easier than with our previous implementation, while also being less invasive by leaving the current logging system in its initial state.

The final goal remains to later on add an entrypoint to the API to list the configuration errors with precise context.

LandryBe:
### What does this PR do?

This PR injects errors in the resource handle by the topology.

Fixes #522

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -71,7 +71,9 @@ func (p *Provider) BuildConfig(t *topology.Topology) *dynamic.Configuration {
 
 	for svcKey, svc := range t.Services {
 		if err := p.buildConfigForService(t, cfg, svc); err != nil {
-			p.logger.Errorf("Unable to build config for service %q: %w", svcKey, err)
+			err = fmt.Errorf("unable to build configuration: %v", err)
+			svc.AddError(err)
+			p.logger.Errorf("Error building dynamic configuration for Service %q: %v", svcKey, err)
 		}
 	}
 
@@ -115,7 +117,10 @@ func (p *Provider) buildConfigForService(t *topology.Topology, cfg *dynamic.Conf
 
 		for _, ttKey := range svc.TrafficTargets {
 			if err := p.buildServicesAndRoutersForTrafficTarget(t, cfg, ttKey, scheme, trafficType, middlewares); err != nil {
-				p.logger.Errorf("Unable to build routers and services for traffic-target %q: %v", ttKey, err)
+				err = fmt.Errorf("unable to build routers and services: %v", err)
+				t.ServiceTrafficTargets[ttKey].AddError(err)
+				p.logger.Errorf("Error building dynamic configuration for TrafficTarget %q: %v", ttKey, err)
+
 				continue
 			}
 		}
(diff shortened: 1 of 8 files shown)


pkg/topology/builder.go
```diff
@@ -89,20 +85,29 @@ func (b *Builder) evaluateService(res *resources, topology *Topology, svc *v1.Se
...
-func (b *Builder) evaluateTrafficTarget(res *resources, topology *Topology, tt *access.TrafficTarget) error {
+func (b *Builder) evaluateTrafficTarget(res *resources, topology *Topology, tt *access.TrafficTarget) {
 	destSaKey := Key{tt.Destination.Name, tt.Destination.Namespace}
 
 	sources := b.buildTrafficTargetSources(res, topology, tt)
 
 	specs, err := b.buildTrafficTargetSpecs(res, tt)
 	if err != nil {
-		return fmt.Errorf("unable to build Specs: %w", err)
+		err = fmt.Errorf("unable to build spec for TrafficTarget %q: %v", Key{tt.Name, tt.Namespace}, err)
```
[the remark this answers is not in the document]
LandryBe: The function doesn't return error anymore, but we need to put a return to stop here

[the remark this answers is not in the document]
Ullaakut: Why does it no longer return an error?

pkg/provider/provider.go
```diff
@@ -170,9 +166,8 @@ func (p *Provider) buildServicesAndRoutersForService(t *topology.Topology, cfg *
 		for _, svcPort := range svc.Ports {
 			entrypoint, err := p.buildTCPEntrypoint(svc, svcPort.Port)
 			if err != nil {
-				err = fmt.Errorf("unable to build TCP entrypoint for Service %q and port %d: %v", svcKey, svcPort.Port, err)
-				svc.Err = err
-				p.logger.Errorf("Error building dynamic configuration: %v", err)
+				svc.Err = fmt.Errorf("unable to build TCP entrypoint  and port %d: %v", svcPort.Port, err)
```
skwair, line 169:
```suggestion
				svc.Err = fmt.Errorf("unable to build TCP entrypoint and port %d: %v", svcPort.Port, err)
```

LandryBe:
Investigating strange behavior, please don't merge right now

pkg/provider/provider.go
```diff
@@ -159,7 +166,9 @@ func (p *Provider) buildServicesAndRoutersForService(t *topology.Topology, cfg *
 		for _, svcPort := range svc.Ports {
 			entrypoint, err := p.buildTCPEntrypoint(svc, svcPort.Port)
 			if err != nil {
-				p.logger.Errorf("Unable to build TCP entrypoint for Service %q and port %d: %v", svcKey, svcPort.Port, err)
+				svc.Err = fmt.Errorf("unable to build TCP entrypoint for port %d: %v", svcPort.Port, err)
```
dtomcej, line 169: Here we loop over ports in the service, and if there is an error, we set the error on the service...but what happens of there are 2 ports that cause errors? Do we care if there are more than one error? Should `svc.Err` be a slice that the error gets appended to?

LandryBe: Indeed, actually we are about to change the type of the ```Err``` from ```error``` to ```[]string``` because we realize that it can be an issue if we want to marsh the topology.
It will solve these 2 issues.

Ullaakut:
(There is a conflict on the provider due to the UDP PR I think)

skwair:
The PR has been rebased and the weird bug fixed, it's ready for a final review :ship:

PR #531, opened 2020-04-30
Multiple middleware support & annotations refactoring

jspdown:
### What does this PR do?

This PR removes code duplication around annotations evaluation. It also fixes a bug in the ShadowServiceManager, when the service has an invalid `traffic-type` value. Before this PR, if a service had an invalid traffic-type, it was considered to be an HTTP service. Since the provider doesn't handle it this way, a shadow service was created but not used.
This PR also moves the middleware builder in the annotations package and the provider now accepts it as a dependency. This will allow external projects to reuse the provider while allowing a customization of the middlewares.

This PR fixes a bug which prevented the use of multiple middlewares on the same service.

### How to test it

- `make test-integration`
- `make test`
- Manually: 
  - Deploy a service with a valid traffic-type
  - Deploy a service with no traffic-type
  - Deploy a service with an invalid traffic-type
  - Apply multiple middleware on the same service


pkg/annotations/annotations.go
```diff
@@ -0,0 +1,57 @@
...
+
+// GetTrafficType returns the value of the traffic-type annotation.
+func GetTrafficType(defaultTrafficType string, annotations map[string]string) (string, error) {
+	trafficType, ok := annotations[annotationServiceType]
+
+	if !ok {
+		return defaultTrafficType, nil
+	}
+
+	if trafficType != ServiceTypeHTTP && trafficType != ServiceTypeTCP && trafficType != ServiceTypeUDP {
```
dtomcej, line 37: This should be a switch with fallthrough to a default?

pkg/annotations/annotations.go
```diff
@@ -0,0 +1,57 @@
...
+
+// GetScheme returns the value of the scheme annotation.
+func GetScheme(annotations map[string]string) (string, error) {
+	scheme, ok := annotations[annotationScheme]
+
+	if !ok {
+		return SchemeHTTP, nil
+	}
+
+	if scheme != SchemeHTTP && scheme != SchemeH2c && scheme != SchemeHTTPS {
```
dtomcej, line 52: same here.

pkg/controller/service.go
```diff
@@ -180,20 +199,25 @@ func (s *ShadowServiceManager) cleanupPortMapping(oldUserSvc *corev1.Service, ne
...
 
-	svcMode := s.getServiceMode(svc)
+	trafficType, err := annotations.GetTrafficType(s.defaultTrafficType, svc.Annotations)
+	if err != nil {
+		return []corev1.ServicePort{}, fmt.Errorf("unable to get service traffic-type: %w", err)
+	}
 
 	for i, sp := range svc.Spec.Ports {
-		if !(svcMode == k8s.ServiceTypeUDP && sp.Protocol == corev1.ProtocolUDP) && !(svcMode != k8s.ServiceTypeUDP && sp.Protocol == corev1.ProtocolTCP) {
+		if !(trafficType == annotations.ServiceTypeUDP && sp.Protocol == corev1.ProtocolUDP) && !(trafficType != annotations.ServiceTypeUDP && sp.Protocol == corev1.ProtocolTCP) {
```
dtomcej, line 215: Is there a way to clean this up to make it more readable, or move to a function?

dtomcej: perhaps a switch with the different commented cases? just because there are now a lot of not comparisons

pkg/annotations/middleware.go
```diff
@@ -1,50 +1,52 @@
...
+			return nil, fmt.Errorf("unable to build retry middleware, %q annotation is invalid: %w", annotationRetryAttempts, err)
 		}
 
 		middleware.Retry = &dynamic.Retry{Attempts: attempts}
 	}
 
 	// Build rate-limit middleware.
-	rateLimitAverage, hasRateLimitAverage := svc.Annotations[k8s.AnnotationRateLimitAverage]
-	rateLimitBurst, hasRateLimitBurst := svc.Annotations[k8s.AnnotationRateLimitBurst]
+	rateLimitAverage, hasRateLimitAverage := annotations[annotationRateLimitAverage]
```
kevinpollet, line 38: Why these annotations don't have a corresponding `get` function?
Same for above annotations.

pkg/provider/provider.go
```diff
@@ -6,13 +6,17 @@ import (
...
+	"github.com/containous/maesh/pkg/annotations"
 	"github.com/containous/maesh/pkg/k8s"
 	"github.com/containous/maesh/pkg/topology"
 	"github.com/containous/traefik/v2/pkg/config/dynamic"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 )
 
+// MiddlewareBuilder is capable of building a middleware from service annotations.
+type MiddlewareBuilder func(annotations map[string]string) (*dynamic.Middleware, error)
```
kevinpollet, line 18: Why not keep the previous type signature? 

```suggestion
type MiddlewareBuilder func(*topology.Service) (*dynamic.Middleware, error)
```

jspdown: It removes the dependency between topology and annotations. I tried to find an example of a middleware which wouldn't be based on annotations and I couldn't find one.

pkg/annotations/annotations.go
```diff
@@ -56,10 +65,63 @@ func GetScheme(annotations map[string]string) (string, error) {
...
+// GetRetryAttempts returns the value of the retry-attempts annotation.
+func GetRetryAttempts(annotations map[string]string) (int, error) {
+	retryAttempts, ok := annotations[annotationRetryAttempts]
+	if !ok {
+		return 0, ErrNotFound
+	}
+
+	attempts, err := strconv.Atoi(retryAttempts)
+	if err != nil {
+		return 0, fmt.Errorf("invalid %q value: %w", annotationRetryAttempts, err)
```
Ullaakut, line 83:
```suggestion
		return 0, fmt.Errorf("invalid value %q: %w", annotationRetryAttempts, err)
```

pkg/annotations/annotations.go
```diff
@@ -56,10 +65,63 @@ func GetScheme(annotations map[string]string) (string, error) {
...
+// GetRateLimitBurst returns the value of the rate-limit-burst annotation.
+func GetRateLimitBurst(annotations map[string]string) (int, error) {
+	rateLimitBurst, ok := annotations[annotationRateLimitBurst]
+	if !ok {
+		return 0, ErrNotFound
+	}
+
+	burst, err := strconv.Atoi(rateLimitBurst)
+	if err != nil {
+		return 0, fmt.Errorf("invalid %q value: %w", annotationRateLimitBurst, err)
```
Ullaakut, line 108:
```suggestion
		return 0, fmt.Errorf("invalid value %q: %w", annotationRateLimitBurst, err)
```

pkg/annotations/annotations.go
```diff
@@ -56,10 +65,63 @@ func GetScheme(annotations map[string]string) (string, error) {
...
+// GetRateLimitAverage returns the value of the rate-limit-average annotation.
+func GetRateLimitAverage(annotations map[string]string) (int, error) {
+	rateLimitAverage, ok := annotations[annotationRateLimitAverage]
+	if !ok {
+		return 0, ErrNotFound
+	}
+
+	average, err := strconv.Atoi(rateLimitAverage)
+	if err != nil {
+		return 0, fmt.Errorf("invalid %q value: %w", annotationRateLimitAverage, err)
```
Ullaakut, line 123:
```suggestion
		return 0, fmt.Errorf("invalid value %q: %w", annotationRateLimitAverage, err)
```

pkg/annotations/annotations.go
```diff
@@ -38,7 +47,7 @@ func GetTrafficType(defaultTrafficType string, annotations map[string]string) (s
 	case ServiceTypeTCP:
 	case ServiceTypeUDP:
 	default:
-		return trafficType, fmt.Errorf("traffic-type annotation references an unsupported traffic type %q", trafficType)
+		return trafficType, fmt.Errorf("%q annotation references an unsupported traffic type %q", annotationServiceType, trafficType)
```
Ullaakut, line 50:
```suggestion
		return trafficType, fmt.Errorf("annotation %q references an unsupported traffic type %q", annotationServiceType, trafficType)
```

pkg/annotations/annotations.go
```diff
@@ -56,10 +65,63 @@ func GetScheme(annotations map[string]string) (string, error) {
 	case SchemeH2C:
 	case SchemeHTTPS:
 	default:
-		return scheme, fmt.Errorf("scheme annotation references an unknown scheme %q", scheme)
+		return scheme, fmt.Errorf("%q annotation references an unknown scheme %q", annotationScheme, scheme)
```
Ullaakut, line 68:
```suggestion
		return scheme, fmt.Errorf("annotation %q references an unknown scheme %q", annotationScheme, scheme)
```

pkg/annotations/middleware.go
```diff
@@ -3,65 +3,102 @@ package annotations
...
-			return nil, fmt.Errorf("unable to build rate-limit middleware, %q annotation is invalid: %w", annotationRateLimitAverage, err)
-		}
+	return &middleware, nil
+}
 
-		burst, err := strconv.Atoi(rateLimitBurst)
-		if err != nil {
-			return nil, fmt.Errorf("unable to build rate-limit middleware, %q annotation is invalid: %w", annotationRateLimitBurst, err)
+func buildRetryMiddleware(middleware *dynamic.Middleware, annotations map[string]string) (bool, error) {
+	retryAttempts, errRetryAttempts := GetRetryAttempts(annotations)
```
Ullaakut, line 38: Why not call this `err`? 🤔 🤔 🤔 🤔

pkg/annotations/middleware.go
```diff
@@ -3,65 +3,102 @@ package annotations
...
+}
+
+func buildRateLimitMiddleware(middleware *dynamic.Middleware, annotations map[string]string) (bool, error) {
+	var (
+		rateLimitBurst   int
+		rateLimitAverage int
+		err              error
+	)
+
+	rateLimitBurst, err = GetRateLimitBurst(annotations)
```
Ullaakut, line 61: There's no shadow declaration risk anymore since we're in a separate func.

```suggestion
	rateLimitBurst, err := GetRateLimitBurst(annotations)
```

pkg/annotations/middleware.go
```diff
@@ -3,65 +3,102 @@ package annotations
...
+	middleware.Retry = &dynamic.Retry{Attempts: retryAttempts}
+
+	return true, nil
+}
+
+func buildRateLimitMiddleware(middleware *dynamic.Middleware, annotations map[string]string) (bool, error) {
+	var (
+		rateLimitBurst   int
+		rateLimitAverage int
+		err              error
```
Ullaakut, line 58:
```suggestion
```

pkg/annotations/middleware.go
```diff
@@ -3,65 +3,102 @@ package annotations
...
-			Burst:   int64(burst),
-		}
+	middleware.Retry = &dynamic.Retry{Attempts: retryAttempts}
+
+	return true, nil
+}
+
+func buildRateLimitMiddleware(middleware *dynamic.Middleware, annotations map[string]string) (bool, error) {
+	var (
+		rateLimitBurst   int
```
Ullaakut, line 56:
```suggestion
```

pkg/annotations/middleware.go
```diff
@@ -7,55 +7,51 @@ import (
...
-	if middlewareCounter == 0 {
-		return nil, nil
-	}
-
-	return &middleware, nil
+	return middlewares, nil
 }
 
-func buildRetryMiddleware(middleware *dynamic.Middleware, annotations map[string]string) (bool, error) {
+func buildRetryMiddleware(annotations map[string]string) (*dynamic.Middleware, string, error) {
```
Ullaakut, line 37: Please use named returns (but not naked returns) to make it clearer what this function will return from reading its prototype

pkg/annotations/middleware.go
```diff
@@ -7,55 +7,51 @@ import (
 	"github.com/containous/traefik/v2/pkg/config/dynamic"
 )
 
-type middlewareBuilder func(middleware *dynamic.Middleware, annotations map[string]string) (bool, error)
+type middlewareBuilder func(annotations map[string]string) (*dynamic.Middleware, string, error)
```
Ullaakut, line 10: Please use named returns to make it clearer what this func type will return from reading its prototype

pkg/annotations/middleware.go
jspdown: It was useful when all these middleware were built in the same function. I guess I just forgot the change them back.

pkg/annotations/middleware.go
```diff
@@ -0,0 +1,111 @@
...
+
+		return nil, "", fmt.Errorf("unable to build retry middleware: %w", err)
+	}
+
+	name = "retry"
+	middleware = &dynamic.Middleware{
+		Retry: &dynamic.Retry{Attempts: retryAttempts},
+	}
+
+	return middleware, name, nil
```
kevinpollet, line 53: Why are you using named return values if you are not using them?
I think a `return` is enough here (previous return can be simplified too).

pkg/annotations/middleware.go
```diff
@@ -0,0 +1,111 @@
...
+
+	name = "rate-limit"
+	middleware = &dynamic.Middleware{
+		RateLimit: &dynamic.RateLimit{
+			Burst:   int64(rateLimitBurst),
+			Average: int64(rateLimitAverage),
+		},
+	}
+
+	return middleware, name, nil
```
kevinpollet, line 88: See my previous comment.

pkg/controller/service.go
```diff
@@ -185,20 +204,24 @@ func (s *ShadowServiceManager) cleanupPortMapping(oldUserSvc *corev1.Service, ne
...
 }
 
-func (s *ShadowServiceManager) getShadowServicePorts(svc *corev1.Service) []corev1.ServicePort {
+func (s *ShadowServiceManager) getShadowServicePorts(svc *corev1.Service) ([]corev1.ServicePort, error) {
 	var ports []corev1.ServicePort
 
-	svcMode := s.getServiceMode(svc)
+	trafficType, err := annotations.GetTrafficType(s.defaultTrafficType, svc.Annotations)
+	if err != nil {
+		return []corev1.ServicePort{}, fmt.Errorf("unable to get service traffic-type: %w", err)
```
kevinpollet, line 216:
```suggestion
		return nil, fmt.Errorf("unable to get service traffic-type: %w", err)
```

pkg/controller/service.go
```diff
@@ -273,6 +288,14 @@ func (s *ShadowServiceManager) getMappedPort(stateTable PortMapper, svcName, svc
 	return port, nil
 }
 
+func isPortSuitable(trafficType string, sp corev1.ServicePort) bool {
+	if trafficType == annotations.ServiceTypeUDP {
+		return sp.Protocol == corev1.ProtocolUDP
+	}
+
+	return sp.Protocol == corev1.ProtocolTCP
```
kevinpollet, line 296: I think it's better to return `false` here and to add a case for `ServiceTypeTCP` && `ServiceTypeHTTP`.

jspdown: I did it this way to handle both `SeviceTypeHTTP` and `ServiceTypeTCP`. I can write it stricter like this:

```
if trafficType == annotations.ServiceTypeUDP {
  return return sp.Protocol == corev1.ProtocolUDP
}
if trafficType == annotations.ServiceTypeTCP || trafficType == annotations.ServiceTypeHTTP {
  return sp.Protocol == corev1.ProtocolTCP
}

return false
```

pkg/annotations/middleware.go
jspdown: I followed @Ullaakut suggestion https://github.com/containous/maesh/pull/531#discussion_r419898894

But regarding naked returns, I would prefer not to. It makes the code harder to digest.

PR #538, opened 2020-05-01
Add Cleanup Command

### LINKED ISSUE #487: Uninstall Helm Chart should remove shadow services

dtomcej:
## What does this PR do?

This PR:

- Adds a cleanup command
- Adds cleanup functionality to purge created shadow services from the cluster on helm delete

Fixes #487

### DIFF
--- a/cmd/cleanup/cleanup.go
+++ b/cmd/cleanup/cleanup.go
@@ -1,0 +1,54 @@
+package cleanup
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/containous/maesh/cmd"
+	"github.com/containous/maesh/pkg/cleanup"
+	"github.com/containous/maesh/pkg/k8s"
+	"github.com/containous/traefik/v2/pkg/cli"
+	"github.com/sirupsen/logrus"
+)
+
+// NewCmd builds a new Cleanup command.
+func NewCmd(cConfig *cmd.CleanupConfiguration, loaders []cli.ResourceLoader) *cli.Command {
+	return &cli.Command{
+		Name:          "cleanup",
+		Description:   `Cleanup command.`,
+		Configuration: cConfig,
+		Run: func(_ []string) error {
+			return cleanupCommand(cConfig)
+		},
+		Resources: loaders,
+	}
+}
+
+func cleanupCommand(cConfig *cmd.CleanupConfiguration) error {
+	log := logrus.New()
+
+	log.SetOutput(os.Stdout)
+	log.SetLevel(logrus.InfoLevel)
+
+	if cConfig.Debug {
+		log.SetLevel(logrus.DebugLevel)
+	}
+
+	log.Debugln("Starting maesh cleanup...")
+	log.Debugf("Using masterURL: %q", cConfig.MasterURL)
+	log.Debugf("Using kubeconfig: %q", cConfig.KubeConfig)
+
+	clients, err := k8s.NewClient(log, cConfig.MasterURL, cConfig.KubeConfig)
+	if err != nil {
+		return fmt.Errorf("error building clients: %v", err)
+	}
+
+	c := cleanup.NewCleanup(log, clients)
+
+	err = c.CleanShadowServices()
+	if err != nil {
+		return fmt.Errorf("error encountered during cluster cleanup: %w", err)
+	}
+
+	return nil
+}
--- a/cmd/cleanup/cleanup.go
+++ b/cmd/cleanup/cleanup.go
@@ -28,11 +28,15 @@ func cleanupCommand(cConfig *cmd.CleanupConfiguration) error {
 	log := logrus.New()
 
 	log.SetOutput(os.Stdout)
-	log.SetLevel(logrus.InfoLevel)
 
-	if cConfig.Debug {
-		log.SetLevel(logrus.DebugLevel)
+	logLevelStr := cConfig.LogLevel
+
+	logLevel, err := logrus.ParseLevel(logLevelStr)
+	if err != nil {
+		return err
 	}
+
+	log.SetLevel(logLevel)
 
 	log.Debugln("Starting maesh cleanup...")
 	log.Debugf("Using masterURL: %q", cConfig.MasterURL)
--- a/cmd/cleanup/cleanup.go
+++ b/cmd/cleanup/cleanup.go
@@ -45,7 +45,7 @@ func cleanupCommand(cConfig *cmd.CleanupConfiguration) error {
 
 	c := cleanup.NewCleanup(log, clients)
 
-	if err := c.CleanShadowServices(); err != nil {
+	if err := c.CleanShadowServices(cConfig.Namespace); err != nil {
 		return fmt.Errorf("error encountered during cluster cleanup: %w", err)
 	}
 
--- a/cmd/cleanup/cleanup.go
+++ b/cmd/cleanup/cleanup.go
@@ -40,12 +40,12 @@ func cleanupCommand(cConfig *cmd.CleanupConfiguration) error {
 
 	clients, err := k8s.NewClient(log, cConfig.MasterURL, cConfig.KubeConfig)
 	if err != nil {
-		return fmt.Errorf("error building clients: %wß", err)
+		return fmt.Errorf("error building clients: %w", err)
 	}
 
-	c := cleanup.NewCleanup(log, clients)
+	c := cleanup.NewCleanup(log, clients, cConfig.Namespace)
 
-	if err := c.CleanShadowServices(cConfig.Namespace); err != nil {
+	if err := c.CleanShadowServices(); err != nil {
 		return fmt.Errorf("error encountered during cluster cleanup: %w", err)
 	}
 
--- a/cmd/cleanup/cleanup.go
+++ b/cmd/cleanup/cleanup.go
@@ -29,9 +29,7 @@ func cleanupCommand(cConfig *cmd.CleanupConfiguration) error {
 
 	log.SetOutput(os.Stdout)
 
-	logLevelStr := cConfig.LogLevel
-
-	logLevel, err := logrus.ParseLevel(logLevelStr)
+	logLevel, err := logrus.ParseLevel(cConfig.LogLevel)
 	if err != nil {
 		return err
 	}
--- a/cmd/cleanup/cleanup.go
+++ b/cmd/cleanup/cleanup.go
@@ -15,7 +15,7 @@ import (
 func NewCmd(cConfig *cmd.CleanupConfiguration, loaders []cli.ResourceLoader) *cli.Command {
 	return &cli.Command{
 		Name:          "cleanup",
-		Description:   `Cleanup command.`,
+		Description:   `Removes Maesh shadow services from a Kubernetes cluster.`,
 		Configuration: cConfig,
 		Run: func(_ []string) error {
 			return cleanupCommand(cConfig)
@@ -40,13 +40,12 @@ func cleanupCommand(cConfig *cmd.CleanupConfiguration) error {
 
 	clients, err := k8s.NewClient(log, cConfig.MasterURL, cConfig.KubeConfig)
 	if err != nil {
-		return fmt.Errorf("error building clients: %v", err)
+		return fmt.Errorf("error building clients: %wß", err)
 	}
 
 	c := cleanup.NewCleanup(log, clients)
 
-	err = c.CleanShadowServices()
-	if err != nil {
+	if err := c.CleanShadowServices(); err != nil {
 		return fmt.Errorf("error encountered during cluster cleanup: %w", err)
 	}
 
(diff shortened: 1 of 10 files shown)


helm/chart/maesh/templates/controller/hooks/cleanup-hook.yaml
```diff
@@ -0,0 +1,52 @@
...
+      containers:
+        - name: maesh-cleanup
+          image: {{ include "maesh.image" . | quote }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | default "IfNotPresent" }}
+          args:
+            - "cleanup"
+            {{- if .Values.controller.logging.debug }}
+            - "--debug"
+            {{- end }}
+            - "--namespace=$(POD_NAMESPACE)"
```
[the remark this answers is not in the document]
kevinpollet: Btw it's consistent with https://github.com/containous/maesh/blob/master/helm/chart/maesh/templates/controller/controller-deployment.yaml#L91

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,52 @@
...
+	return &Cleanup{
+		client: client,
+		log:    log,
+	}
+}
+
+// CleanShadowServices deletes all shadow services from the cluster.
+func (c *Cleanup) CleanShadowServices() error {
+	serviceList, err := c.client.GetKubernetesClient().CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{
+		LabelSelector: "app=maesh",
```
kevinpollet, line 34: Are you sure that only shadow services will be returned?
I'm thinking that this could have side effects with the `maesh-controller` service introduced in #525.

cmd/cleanup/cleanup.go
```diff
@@ -0,0 +1,54 @@
...
+	log.Debugf("Using kubeconfig: %q", cConfig.KubeConfig)
+
+	clients, err := k8s.NewClient(log, cConfig.MasterURL, cConfig.KubeConfig)
+	if err != nil {
+		return fmt.Errorf("error building clients: %v", err)
+	}
+
+	c := cleanup.NewCleanup(log, clients)
+
+	err = c.CleanShadowServices()
```
Ullaakut, line 48:
```suggestion
	if err := c.CleanShadowServices(); err != nil {
```

cmd/cleanup/cleanup.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
```
Ullaakut, line 49:
```suggestion
```

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Cleaner is an interface for the cleanup methods.
+type Cleaner interface {
+	CleanShadowServices() error
+}
+
+// Ensure the Prepare fits the Preparer interface
```
Ullaakut, line 14: Copy paste engineer! :p 

```suggestion
// Ensure the Cleanup fits the Cleaner interface
```

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,52 @@
...
+func (c *Cleanup) CleanShadowServices() error {
+	serviceList, err := c.client.GetKubernetesClient().CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{
+		LabelSelector: "app=maesh",
+	})
+	if len(serviceList.Items) == 0 {
+		// No services found, return without error.
+		return nil
+	}
+
+	if err != nil {
```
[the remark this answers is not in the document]
Ullaakut: Yep, we should check for errors before checking for an empty response, otherwise in the case of a wrong kubeconfig or wrong kubernetes API address for example, users might think that there is nothing to be cleaned and it was successful, while they just provided invalid parameters.

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,52 @@
...
+		log:    log,
+	}
+}
+
+// CleanShadowServices deletes all shadow services from the cluster.
+func (c *Cleanup) CleanShadowServices() error {
+	serviceList, err := c.client.GetKubernetesClient().CoreV1().Services(metav1.NamespaceAll).List(metav1.ListOptions{
+		LabelSelector: "app=maesh",
+	})
+	if len(serviceList.Items) == 0 {
```
Ullaakut, line 36: Can't `serviceList` be nil?

cmd/cleanup/cleanup.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/containous/maesh/pkg/k8s"
+	"github.com/containous/traefik/v2/pkg/cli"
+	"github.com/sirupsen/logrus"
+)
+
+// NewCmd builds a new Cleanup command.
+func NewCmd(cConfig *cmd.CleanupConfiguration, loaders []cli.ResourceLoader) *cli.Command {
+	return &cli.Command{
+		Name:          "cleanup",
+		Description:   `Cleanup command.`,
```
Ullaakut, line 18:
```suggestion
		Description:   `Removes Maesh shadow services from a Kubernetes cluster.`,
```

WDYT?

helm/chart/maesh/templates/controller/hooks/cleanup-hook.yaml
```diff
@@ -0,0 +1,52 @@
...
+metadata:
+  name: maesh-cleanup
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Release.Name | quote}}
+    chart: {{ include "maesh.chartLabel" . | quote }}
+    release: {{ .Release.Name | quote}}
+    heritage: {{ .Release.Service | quote}}
+  annotations:
+    "helm.sh/hook": "pre-delete"
```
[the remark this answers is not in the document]
dtomcej: Do you think there would be an issue with service accounts? Since the service account we use to run the cleanup is managed by helm.

Or do you think we should create its own service account? idk.

helm/chart/maesh/templates/controller/hooks/cleanup-hook.yaml
```diff
@@ -0,0 +1,52 @@
...
+    "helm.sh/hook-delete-policy": "hook-succeeded,before-hook-creation"
```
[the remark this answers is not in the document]
dtomcej: Hook failed is excluded so that yes, we can see whats going on if this fails.

pkg/cleanup/cleanup.go
dtomcej: You are correct, we should use a new label for shadow services to prevent this.

helm/chart/maesh/templates/controller/hooks/cleanup-hook.yaml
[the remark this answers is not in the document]
kevinpollet: I think that the running controller could create a shadow service just after the cleanup. Btw, it is a corner case.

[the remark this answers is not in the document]
dtomcej: unless we bundle an RBAC with the hook? and then that way it can remove services after the controller has been deleted?

pkg/cleanup/cleanup.go
kevinpollet: So if we use a new label, it will not work for previous versions of maesh, right?

dtomcej: I was going to add an additional label. You are correct, the cleanup would not work on older versions without the label.

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,52 @@
+package cleanup
+
+import (
+	"github.com/containous/maesh/pkg/k8s"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Cleaner is an interface for the cleanup methods.
+type Cleaner interface {
```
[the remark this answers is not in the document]
dtomcej: No, not really.

helm/chart/maesh/templates/controller/hooks/cleanup-hook.yaml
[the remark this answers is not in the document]
dtomcej: That is a callback to when we forced the install of maesh to a different namespace than the release was tied to.

This ensured the namespace used was the namespace the pod was running in.

pkg/cleanup/cleanup.go
dtomcej: Only the best.

pkg/cleanup/cleanup.go
dtomcej: I don't think so, its not a pointer:

https://godoc.org/k8s.io/api/core/v1#ServiceList

although I guess an empty slice is nil isn't it? but wouldn't that also still work with the range?

dtomcej: I have removed this check since we just iterate over the items after the error check.

pkg/cleanup/cleanup.go
dtomcej: The other option would be to have a more complex service filter that would ignore the controller and API services created by helm.

This would also be backwards compatible.

@kevinpollet which method do you think you would want to see? (new label vs more involved service search)

jspdown: I agree with what @kevinpollet said. Additionally, could we specify the maesh namespace instead of `metav1.NamespaceAll`?

SantoDE: I think a new label would be good. Of course, being backwards compatible is preferred but we don't have a clean up as of now. So only, for new meshes the new feature will work. It's not like we're breaking something, it's like "it will not work".

Maybe we could even reconcile that and add the new labels to existing shadow services? @dtomcej

pkg/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"os"
+	"testing"
+
+	"github.com/containous/maesh/pkg/k8s"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestNewCleanup(t *testing.T) {
```
kevinpollet, line 14:
```suggestion
func TestCleanup_New(t *testing.T) {
```

pkg/cleanup/cleanup_test.go
```diff
@@ -0,0 +1,53 @@
...
+	log := logrus.New()
+
+	log.SetOutput(os.Stdout)
+	log.SetLevel(logrus.DebugLevel)
+
+	cln := NewCleanup(log, clientMock)
+	assert.NotNil(t, cln)
+}
+
+func TestCleanShadowServices(t *testing.T) {
```
kevinpollet, line 28:
```suggestion
func TestCleanup_CleanShadowServices(t *testing.T) {
```

pkg/cleanup/cleanup.go
```diff
@@ -0,0 +1,39 @@
...
+// NewCleanup returns an initialized cleanup object.
+func NewCleanup(log logrus.FieldLogger, client k8s.Client) *Cleanup {
+	return &Cleanup{
+		client: client,
+		log:    log,
+	}
+}
+
+// CleanShadowServices deletes all shadow services from the cluster.
+func (c *Cleanup) CleanShadowServices(namespace string) error {
```
kevinpollet, line 24: To be consistent, maybe the `namespace` could be a `Cleanup` struct field?

cmd/cleanup/cleanup.go
```diff
@@ -0,0 +1,53 @@
...
+		log.SetLevel(logrus.DebugLevel)
+	}
+
+	log.Debugln("Starting maesh cleanup...")
+	log.Debugf("Using masterURL: %q", cConfig.MasterURL)
+	log.Debugf("Using kubeconfig: %q", cConfig.KubeConfig)
+
+	clients, err := k8s.NewClient(log, cConfig.MasterURL, cConfig.KubeConfig)
+	if err != nil {
+		return fmt.Errorf("error building clients: %wß", err)
```
kevinpollet, line 43:
```suggestion
		return fmt.Errorf("error building clients: %w", err)
```

dtomcej: No clue how that happened

PR #551, opened 2020-05-05
Add destination namespace check in indexer

### LINKED ISSUE #546: TrafficTarget deployed in the default namespace does not work
### What did you do?

Deployed a `TrafficTarget` in the `default` namespace to allow `HTTP` traffic from a client pod to a `whoami` service. The same deployment works if the `TrafficTarget` is not deployed in the `default` namespace.

`HTTP 200 OK`

### What did you see instead?

`HTTP 403 Forbidden`

### What is your environment & configuration?

ACL mode enabled.

### If applicable, please paste the yaml objects required to reproduce your issue

- **acl.yml**

```yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: client-account

---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: whoami-account

---
kind: HTTPRouteGroup
metadata:
  name: http-routes
matches:
  - name: everything
    pathRegex: ".*"
    methods: ["*"]

---
kind: TrafficTarget
metadata:
  name: traffic-target
destination:
  kind: ServiceAccount
  name: whoami-account
  port: "80"
specs:
  - kind: HTTPRouteGroup
    name: http-routes
sources:
  - kind: ServiceAccount
    name: client-account
```

- **client.yml**

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: client
spec:
  replicas: 1
  selector:
    matchLabels:
      app: client
  template:
    metadata:
      labels:
        app: client
    spec:
      serviceAccountName: client-account
      containers:
        - name: client
          image: giantswarm/tiny-tools:3.9
          command:
            - "sleep"
            - "3600"
```

- **whoami.yml**

```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: whoami
spec:
  selector:
    app: whoami
  ports:
    - port: 80

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whoami
spec:
  selector:
    matchLabels:
      app: whoami
  template:
    metadata:
      labels:
        app: whoami
    spec:
      serviceAccountName: whoami-account
      containers:
        - name:  whoami
          image: containous/whoami
```

dtomcej:
## What does this PR do?

This PR:

- Adds a check for an empty destination namespace in traffictargets during load
- Adds test to ensure empty destination namespaces are filled with the namespace of the traffictarget.

Fixes #546

## Additional Notes

We may want to look at a "validate resource" method in the future, but I think for now it can be avoided as there is not a lot of validation required.

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -141,7 +141,6 @@ By default, all traffic is denied so we need to grant access to clients to our a
 
 !!! Note "TrafficTarget Source & Destination"
     Please note that TrafficTarget is a namespaced resource.
-    It is important to ensure that the destination namespace is populated, as kubernetes will not autopopulate this field for you.
     If the destination namespace is not populated, the TrafficTarget namespace will be used as the destination namespace. 
 
 ```yaml
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -141,7 +141,8 @@ By default, all traffic is denied so we need to grant access to clients to our a
 
 !!! Note "TrafficTarget Source & Destination"
     Please note that TrafficTarget is a namespaced resource.
-    If the destination namespace is not populated, the TrafficTarget namespace will be used as the destination namespace. 
+    If the destination namespace is not populated, the TrafficTarget namespace will be used as the destination namespace.
+    The source namespace must be populated, as it cannot be inferred.
 
 ```yaml
 ---
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -139,9 +139,10 @@ Other types of route groups and detailed information are available [in the speci
 
 By default, all traffic is denied so we need to grant access to clients to our application. This is done by defining a `TrafficTarget`.
 
-??? Note "TrafficTarget Source & Destination"
-    Please note that TrafficTarget is a namespaced resource. Therefore, the source and the destination namespace needs to be explicitly defined.
-    If you do not define a destination namespace, the TrafficTarget namespace will be used as the destination namespace.
+!!! Note "TrafficTarget Source & Destination"
+    Please note that TrafficTarget is a namespaced resource.
+    It is important to ensure that the destination namespace is populated, as kubernetes will not autopopulate this field for you.
+    If the destination namespace is not populated, the TrafficTarget namespace will be used as the destination namespace. 
 
 ```yaml
 ---
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -141,6 +141,7 @@ By default, all traffic is denied so we need to grant access to clients to our a
 
 ??? Note "TrafficTarget Source & Destination"
     Please note that TrafficTarget is a namespaced resource. Therefore, the source and the destination namespace needs to be explicitly defined.
+    If you do not define a destination namespace, the TrafficTarget namespace will be used as the destination namespace.
 
 ```yaml
 ---
--- a/pkg/topology/builder.go
+++ b/pkg/topology/builder.go
@@ -35,7 +35,7 @@ type Builder struct {
 func (b *Builder) Build(ignoredResources mk8s.IgnoreWrapper) (*Topology, error) {
 	topology := NewTopology()
 
-	res, err := b.LoadResources(ignoredResources)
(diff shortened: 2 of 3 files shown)


jspdown:
@dtomcej Thanks for the PR. Initially I was thinking about doing this in the `indexSMIResources`. The sooner is the better I think.

SantoDE:
Just a general question first:

I see that this takes an empty namespace as default, which I think is good. However, how do we handle the case if Source or Destination don't have a namespace set. Do we infer the Namespace of the TrafficTarget?

dtomcej:
Good call @SantoDE. I checked the CRD, and both source _and_ destination have namespace as an optional field.

I will assume that no namespace defined would be default, but that could easily be changed in the future. For the time being, we will use default, so that code functions properly.

I have opened an issue: https://github.com/servicemeshinterface/smi-spec/issues/162 that should hopefully answer this question.

SantoDE:
Thanks @dtomcej.

Seeing that the issue is mostly answered by Thomas already, I can live with that. We should just document the behaviour tough :-)

dtomcej:
As answered in the SMI spec issue above, we will have the namespace inherited from the TT object itself.

docs/content/configuration.md
```diff
@@ -141,6 +141,7 @@ By default, all traffic is denied so we need to grant access to clients to our a
 
 ??? Note "TrafficTarget Source & Destination"
     Please note that TrafficTarget is a namespaced resource. Therefore, the source and the destination namespace needs to be explicitly defined.
+    If you do not define a destination namespace, the TrafficTarget namespace will be used as the destination namespace.
```
kevinpollet, line 144: I think that we should also reword the previous sentence because we're saying that the `destination` namespace is required, which is not valid. Also, I think that this topic can confuse users, so maybe this note could be expanded by default?

pkg/topology/builder.go
```diff
@@ -722,6 +722,11 @@ func (r *resources) indexSMIResources(ignoredResources mk8s.IgnoreWrapper, tts [
 			continue
 		}
 
+		// If the destination namepace is empty or blank, set it to the trafficTarget namespace.
+		if len(trafficTarget.Destination.Namespace) == 0 {
```
kevinpollet, line 726:
```suggestion
		if trafficTarget.Destination.Namespace == "" {
```

docs/content/configuration.md
```diff
@@ -139,8 +139,10 @@ Other types of route groups and detailed information are available [in the speci
 
 By default, all traffic is denied so we need to grant access to clients to our application. This is done by defining a `TrafficTarget`.
 
-??? Note "TrafficTarget Source & Destination"
-    Please note that TrafficTarget is a namespaced resource. Therefore, the source and the destination namespace needs to be explicitly defined.
+!!! Note "TrafficTarget Source & Destination"
+    Please note that TrafficTarget is a namespaced resource.
+    It is important to ensure that the destination namespace is populated, as kubernetes will not autopopulate this field for you.
```
kevinpollet, line 144:
```suggestion
    It is important to ensure that the source namespace is populated, as kubernetes will not autopopulate this field for you.
```

dtomcej: The entire source field is not mandatory. Only destination is mandatory

kevinpollet: But a source namespace is mandatory?

For me, It's a bit strange to say that it is `important to ensure that the destination namespace is populated` and in the next sentence that `if the destination namespace is not populated, the TrafficTarget namespace will be used`

dtomcej: nope source namespace is not mandatory.

I think the first sentence can be removed.

kevinpollet: hmm, I'm missing something here. 

So, if the source namespace is not mandatory, is the current code working if the source namespace is not populated? 
I'm thinking that we will have the same problem, right?

dtomcej: You don't have to specify a source at all.

No source = All sources allowed.

I'm not sure we handle that correctly at the time being. We should open a new issue for correct source handling in cases of missing source field.

dtomcej: 👍

PR #554, opened 2020-05-06
Rebuild dynamic configuration on pod events

### LINKED ISSUE #513: Pod events listener
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

<details>
  <summary>Deploy a service with its pods and a traffic-target authorizing <b>client</b> to hit <b>server-a</b></summary>

```
apiVersion: v1
kind: Namespace
metadata:
  name: test

---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: client
  namespace: test

---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: server-a
  namespace: test

---
kind: Deployment
apiVersion: apps/v1
metadata:
  name: server-a
  namespace: test
spec:
  selector:
    matchLabels:
      app: server-a
  replicas: 2
  template:
    metadata:
      labels:
        app: server-a
    spec:
      serviceAccountName: server-a
      containers:
        - name: server-a
          image: containous/whoami:v1.0.1
          imagePullPolicy: Always

---
apiVersion: v1
kind: Service
metadata:
  name: server-a
  namespace: test
  labels:
    app: server-a
spec:
  type: ClusterIP
  ports:
    - port: 8080
      name: server-a
      targetPort: 80
  selector:
    app: server-a

---
kind: HTTPRouteGroup
metadata:
  name: http-route-group-api
  namespace: test
matches:
  - name: api
    pathRegex: /api
    methods: ["GET"]

---
kind: TrafficTarget
metadata:
  name: client-to-server-a
  namespace: test
sources:
  - kind: ServiceAccount
    name: client
    namespace: test
destination:
  kind: ServiceAccount
  name: server-a
  namespace: test
specs:
  - kind: HTTPRouteGroup
    name: http-route-group-api
    matches:
      - api
```

</details>

<details>
<summary>Then deploy a pod with the <b>client</b> ServiceAccount:
</summary>

```
apiVersion: v1
kind: Pod
metadata:
  name: client
  namespace: test
spec:
  serviceAccountName: client
  containers:
    - name: client
      image: giantswarm/tiny-tools:3.9
      command:
        - "sleep"
        - "36000"
```

</details>

I was expected to see the client pod creation and update events being interpreted by the controller.

### What did you see instead?

Nothing, these events have been skipped. 

```
case *corev1.Pod:
		if !isMeshPod(obj) {
			// We don't track updates of user pods, updates are done through endpoints.
			return
		}
                ...
```

This comment is partially true. In the context of a TrafficTarget we can definitely keep track of destination pods using endpoints. But for sources, we have to rely on pods themselves. A source pod may not have been exposed behind a Service. 

- k3d: v1.6.0
- k3s: v1.17.2-k3s1
- maesh: master

kevinpollet:
## What does this PR do?

This PR:

- Removes some unused properties and methods.
- Removes the forced configuration rebuild on `mesh` pods events.
- Triggers a configuration rebuild when pod events are received in ACL mode.

Fixes #513.

## How to test it

- Build and install Maesh with ACL mode enabled.
- Deploy the attached [resources](https://github.com/containous/maesh/files/4586145/issue-513.zip).
- Remove/Add the client resource to trigger dynamic config updates (which does not work well on master).

## Additional Notes

I've added a check to the `onAdd`, `onUpdate` and `onDelete` handler methods to make sure that we're not rebuilding the dynamic conf if the event should be ignored. Previously, only the service or endpoint events were filtered but we should do the same for `TrafficTarget` and so on.

As `proxy` nodes are now pulling the configuration from the `controller`, I've removed the unnecessary configuration rebuild when a `mesh` pod is added or deleted. `mesh` pod events are now filtered by the new `isIgnoredResource` method because we're ignoring the `maesh` app by default.

Last but not least, the `Handler` does not ignore pod events anymore in ACL mode. If I'm not wrong it's unnecessary to listen for pod events in non ACL mode.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -20,10 +20,12 @@ import (
 	splitlister "github.com/servicemeshinterface/smi-sdk-go/pkg/gen/client/split/listers/split/v1alpha2"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/informers"
 	listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
 )
 
 // PortMapper is capable of storing and retrieving a port mapping for a given service.
@@ -64,9 +66,9 @@ type Config struct {
 // Controller hold controller configuration.
 type Controller struct {
 	cfg               Config
-	handler           *Handler
+	handler           cache.ResourceEventHandler
 	serviceManager    ServiceManager
-	configRefreshChan chan string
+	configRefreshChan chan struct{}
 	provider          *provider.Provider
 	ignoredResources  k8s.IgnoreWrapper
 	tcpStateTable     PortMapper
@@ -136,8 +138,11 @@ func (c *Controller) init() {
 	c.serviceManager = NewShadowServiceManager(c.logger, c.ServiceLister, c.cfg.Namespace, c.tcpStateTable, c.udpStateTable, c.cfg.DefaultMode, c.cfg.MinHTTPPort, c.cfg.MaxHTTPPort, c.clients.GetKubernetesClient())
 
 	// configRefreshChan is used to trigger configuration refreshes.
-	c.configRefreshChan = make(chan string)
-	c.handler = NewHandler(c.logger, c.ignoredResources, c.serviceManager, c.configRefreshChan)
+	c.configRefreshChan = make(chan struct{})
+	c.handler = cache.FilteringResourceEventHandler{
+		FilterFunc: c.isWatchedResource,
+		Handler:    NewHandler(c.logger, c.serviceManager, c.configRefreshChan),
+	}
 
 	// Create listers and register the event handler to informers that are not ACL related.
 	c.PodLister = c.kubernetesFactory.Core().V1().Pods().Lister()
@@ -308,4 +313,16 @@ func (c *Controller) createMeshServices() error {
 	}
 
 	return nil
+}
+
+// isWatchedResource returns true if the given resource is not ignored, false otherwise.
+func (c *Controller) isWatchedResource(obj interface{}) bool {
+	accessor, err := meta.Accessor(obj)
+	if err != nil {
+		return false
+	}
+
+	pMeta := meta.AsPartialObjectMetadata(accessor)
+
+	return !c.ignoredResources.IsIgnored(pMeta.ObjectMeta)
 }
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -182,7 +182,6 @@ func (c *Controller) init() {
 		MaxHTTPPort:        c.cfg.MaxHTTPPort,
 		ACL:                c.cfg.ACLEnabled,
 		DefaultTrafficType: c.cfg.DefaultMode,
-		MaeshNamespace:     c.cfg.Namespace,
 	}
 
 	c.provider = provider.New(c.tcpStateTable, c.udpStateTable, annotations.BuildMiddlewares, providerCfg, c.logger)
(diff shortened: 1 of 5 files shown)


pkg/controller/handler.go
```diff
@@ -4,72 +4,70 @@ import (
 	"github.com/containous/maesh/pkg/k8s"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 )
 
+// configMessageChanRebuild rebuild.
+const configMessageChanRebuild = "rebuild"
```
jspdown, line 11: We could get rid of this and change the `configRefreshChan` from a `string` into `struct{}`. We are not interested in any information we receive from this chan. WDYT?

pkg/controller/handler.go
```diff
@@ -4,72 +4,70 @@ import (
...
-	switch obj := obj.(type) {
-	case *corev1.Service:
-		if h.ignored.IsIgnored(obj.ObjectMeta) {
-			return
-		}
+	if h.isIgnoredResource(obj) {
+		return
+	}
 
+	// If the created object is a service we should create a corresponding shadow service.
```
jspdown, line 37:
```suggestion
	// If the created object is a service we have to create a corresponding shadow service.
```

pkg/controller/handler.go
```diff
@@ -4,72 +4,70 @@ import (
...
-		return
-	case *corev1.Pod:
-		if !isMeshPod(obj) {
-			return
-		}
 	}
 
 	// Trigger a configuration rebuild.
-	h.configRefreshChan <- k8s.ConfigMessageChanRebuild
+	h.configRefreshChan <- configMessageChanRebuild
```
jspdown, line 45: Some remark here. I'm almost sure we can send a `struct{}{}`

pkg/controller/handler.go
```diff
@@ -4,72 +4,70 @@ import (
...
-	// Assert the type to an object to pull out relevant data.
-	switch obj := newObj.(type) {
-	case *corev1.Service:
-		if h.ignored.IsIgnored(obj.ObjectMeta) {
-			return
-		}
+	isOldIgnored := h.isIgnoredResource(oldObj)
+	isNewIgnored := h.isIgnoredResource(newObj)
+
+	switch {
```
jspdown, line 53: That's a nice addition! 
I discovered recentely that kubernetes already implements something like this: `cache.FilteringResourceEventHandler`.
It's implementation is really close to what you did: https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L264
WDYT of using this helper instead?

pkg/controller/controller.go
```diff
@@ -182,7 +182,6 @@ func (c *Controller) init() {
 		MaxHTTPPort:        c.cfg.MaxHTTPPort,
 		ACL:                c.cfg.ACLEnabled,
 		DefaultTrafficType: c.cfg.DefaultMode,
-		MaeshNamespace:     c.cfg.Namespace,
```
dtomcej, line 185: Why this removal?

kevinpollet: If I'm not wrong this property is not used in the provider.

dtomcej: it looks like you are right!

pkg/controller/handler.go
```diff
@@ -1,75 +1,46 @@
...
-		}
 	}
 
 	// Trigger a configuration rebuild.
-	h.configRefreshChan <- k8s.ConfigMessageChanRebuild
+	h.configRefreshChan <- configMessageChanRebuild
 }
 
-// OnUpdate executed when an object is updated.
+// OnUpdate is called when an object is updated and ensures that the proper handler is called depending on the filter matches.
```
Ullaakut, line 40:
```suggestion
// OnUpdate is called when an object is updated.
```

There is no filtering anymore

pkg/controller/handler.go
```diff
@@ -1,75 +1,46 @@
 package controller
 
 import (
-	"github.com/containous/maesh/pkg/k8s"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 )
 
+// configMessageChanRebuild rebuild.
+var configMessageChanRebuild = struct{}{}
```
Ullaakut, line 9: WDYT of just sending a `struct{}{}` but not having a variable for it? If the channel names are explicit enough, we should not need to have a variable for this.

PR #564, opened 2020-05-11
Upgrade golangci-lint and refactor to meet proper linting settings

### LINKED ISSUE #550: Upgrade golangci-lint version to v1.26.0
## Feature Request

Everything is in the title.

dtomcej:
## What does this PR do?

This PR:

- Updates golangci-lint to v1.27.0
- Resets gocognit setting to proper value
- Refactors methods to pass proper linting

Fixes #550

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ WORKDIR /go/src/github.com/containous/maesh
 RUN curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh | sh
 
 # Download golangci-lint binary to bin folder in $GOPATH
-RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin v1.23.8
+RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin v1.26.0
 
 ENV GO111MODULE on
 COPY go.mod go.sum ./
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ WORKDIR /go/src/github.com/containous/maesh
 RUN curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh | sh
 
 # Download golangci-lint binary to bin folder in $GOPATH
-RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin v1.26.0
+RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin v1.27.0
 
 ENV GO111MODULE on
 COPY go.mod go.sum ./
--- a/cmd/proxy/proxy.go
+++ b/cmd/proxy/proxy.go
@@ -133,16 +133,7 @@ func setupServer(proxyConfiguration *cmd.ProxyConfiguration) (*server.Server, er
 	managerFactory := service.NewManagerFactory(proxyConfiguration.Configuration, routinesPool, metricsRegistry)
 	routerFactory := server.NewRouterFactory(proxyConfiguration.Configuration, managerFactory, tlsManager, chainBuilder)
 
-	var defaultEntryPoints []string
-
-	for name, cfg := range proxyConfiguration.Configuration.EntryPoints {
-		protocol, _ := cfg.GetProtocol()
-		if protocol != "udp" && name != static.DefaultInternalEntryPointName {
-			defaultEntryPoints = append(defaultEntryPoints, name)
-		}
-	}
-
-	sort.Strings(defaultEntryPoints)
+	defaultEntryPoints := createAndSortDefaultEntrypoints(proxyConfiguration.Configuration.EntryPoints)
 
 	watcher := server.NewConfigurationWatcher(
 		routinesPool,
@@ -175,6 +166,21 @@ func setupServer(proxyConfiguration *cmd.ProxyConfiguration) (*server.Server, er
 	})
 
 	return server.NewServer(routinesPool, serverEntryPointsTCP, serverEntryPointsUDP, watcher, chainBuilder, accessLog), nil
+}
+
+func createAndSortDefaultEntrypoints(eps static.EntryPoints) []string {
+	var defaultEntryPoints []string
+
+	for name, cfg := range eps {
+		protocol, _ := cfg.GetProtocol()
+		if protocol != "udp" && name != static.DefaultInternalEntryPointName {
+			defaultEntryPoints = append(defaultEntryPoints, name)
+		}
+	}
+
+	sort.Strings(defaultEntryPoints)
+
+	return defaultEntryPoints
 }
 
 func switchRouter(routerFactory *server.RouterFactory, serverEntryPointsTCP server.TCPEntryPoints, serverEntryPointsUDP server.UDPEntryPoints) func(conf dynamic.Configuration) {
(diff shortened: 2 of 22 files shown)


pkg/topology/builder.go
```diff
@@ -302,16 +309,7 @@ func (b *Builder) populateTrafficSplitsAuthorizedIncomingTraffic(topology *Topol
...
-		for _, loopTs := range tss {
-			for i, ts := range svc.TrafficSplits {
-				if ts == loopTs {
-					svc.TrafficSplits = append(svc.TrafficSplits[:i], svc.TrafficSplits[i+1:]...)
-					break
-				}
-			}
-		}
-	}
+	removeLoopTrafficSplits(loopDetected)
```
kevinpollet, line 312:
```suggestion
	removeTrafficSplitsLoop(loopDetected)
```

Ullaakut:
@kevinpollet Harold is off for now, so I'll take a look

pkg/controller/service.go
```diff
@@ -318,3 +309,17 @@ func parseKubernetesServerVersion(kubeClient kubernetes.Interface) (major, minor
 
 	return major, minor
 }
+
+func matchPort(old corev1.ServicePort, newUserSvc *corev1.Service) bool {
```
Ullaakut, line 313: Why not pass the port directly instead of the `corev1.ServicePort`?

pkg/k8s/client_mock.go
```diff
@@ -91,82 +91,93 @@ func NewClientMock(stopCh <-chan struct{}, path string, smi bool) *ClientMock {
...
+	return c
+}
+
+// startInformers waits for the kubernetes core informers to start and sync.
+func (c *ClientMock) startInformers(stopCh <-chan struct{}) {
+	c.informerFactory.Start(stopCh)
+
+	for t, ok := range c.informerFactory.WaitForCacheSync(stopCh) {
+		if !ok {
+			fmt.Printf("timed out waiting for controller caches to sync: %s", t.String())
```
Ullaakut, line 152: Maybe not in this PR, but it could make sense to make the mock take a `t *testing.T` and to `t.Log` instead of `Printf`ing

pkg/prepare/prepare.go
```diff
@@ -373,6 +373,22 @@ func (p *Prepare) StartInformers(acl bool) error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 
+	err := p.startBaseInformers(ctx, stopCh)
```
Ullaakut, line 376: Can be inlined

pkg/prepare/prepare.go
```diff
@@ -373,6 +373,22 @@ func (p *Prepare) StartInformers(acl bool) error {
...
+	if err != nil {
+		return err
+	}
+
+	if acl {
+		err = p.startACLInformers(ctx, stopCh)
```
Ullaakut, line 382: Can be inlined

pkg/provider/rule.go
```diff
@@ -58,6 +37,42 @@ func buildTrafficTargetRule(tt *topology.ServiceTrafficTarget) string {
 	return strings.Join(orRules, " || ")
 }
 
+func buildPathFilter(match *specs.HTTPMatch, matchParts []string) []string {
+	if match.PathRegex == "" {
+		return matchParts
+	}
+
+	pathRegex := match.PathRegex
+
```
Ullaakut, line 46:
```suggestion
```

pkg/topology/builder.go
```diff
@@ -745,3 +779,16 @@ func mapCopy(m map[Key]struct{}) map[Key]struct{} {
 
 	return cpy
 }
+
+func removeLoopTrafficSplits(loopDetected map[*Service][]Key) {
```
Ullaakut, line 783: The variable naming in this func is quite confusing to me

cmd/proxy/proxy.go
```diff
@@ -177,6 +168,21 @@ func setupServer(proxyConfiguration *cmd.ProxyConfiguration) (*server.Server, er
...
+	var defaultEntryPoints []string
+
+	for name, cfg := range eps {
+		protocol, _ := cfg.GetProtocol()
+		if protocol != "udp" && name != static.DefaultInternalEntryPointName {
+			defaultEntryPoints = append(defaultEntryPoints, name)
+		}
+	}
+
+	sort.Strings(defaultEntryPoints)
```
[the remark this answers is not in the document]
dtomcej: We sort here so that it creates a stable configuration that prevents order of elements from triggering a configuration refresh.

pkg/prepare/prepare.go
```diff
@@ -49,7 +49,7 @@ type Preparer interface {
 	ConfigureKubeDNS() error
 }
 
-// Ensure the Prepare fits the Preparer interface
+// Ensure the Prepare fits the Preparer interface.
 var _ Preparer = (*Prepare)(nil)
```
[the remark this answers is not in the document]
dtomcej: For use via the EE integration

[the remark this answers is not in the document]
dtomcej: For use in other projects. They use the interface instead of having to worry about the behind-the scenes code.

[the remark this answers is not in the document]
kevinpollet: I'm not sure that the `Preparer` interface is needed because the `Preparer` is always used as-is by the commands and never in a context where we could offer an extension point.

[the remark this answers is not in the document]
dtomcej: I implemented it so that if we modified the Preparer object, that we would not break any integrations with other projects, as they would use the interface.

integration/integration_test.go
```diff
@@ -509,6 +485,34 @@ func matchesConfig(wantConfig string, buf *bytes.Buffer) try.ResponseCondition {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read response body: %s", err)
+	}
+
+	if err = res.Body.Close(); err != nil {
+		return nil, err
+	}
+
+	var obtained dynamic.Configuration
+
```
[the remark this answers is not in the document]
dtomcej: Only one cuddle statement allowed before if statements.

pkg/k8s/client_mock.go
dtomcej: This is a great idea, I will open a new issue for this.

dtomcej: https://github.com/containous/maesh/issues/578

pkg/prepare/prepare.go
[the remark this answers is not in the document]
kevinpollet: I'm not sure to get the point but nevermind that's out of the PR scope. For me, it's the same as using the public methods from the struct, which defines the API. Interfaces express behaviour which should be defined on the [consumer side](https://github.com/golang/go/wiki/CodeReviewComments#interfaces).

pkg/provider/provider.go
```diff
@@ -147,67 +128,122 @@ func (p *Provider) buildConfigForService(t *topology.Topology, cfg *dynamic.Conf
...
+func (p *Provider) buildConfigRoutersAndServices(t *topology.Topology, cfg *dynamic.Configuration, svc *topology.Service, scheme, trafficType string, middlewareKeys []string) error {
+	err := p.buildServicesAndRoutersForService(t, cfg, svc, scheme, trafficType, middlewareKeys)
+	if err != nil {
+		return fmt.Errorf("unable to build routers and services: %w", err)
+	}
+
+	return nil
+}
+
+func (p *Provider) buildACLConfigRoutersAndServices(t *topology.Topology, cfg *dynamic.Configuration, svc *topology.Service, scheme, trafficType string, middlewareKeys []string) {
```
kevinpollet, line 158: I'm a bit surprised that this method does not return an error. I think it should also return an error to indicate to the caller that something is wrong. Then, the caller could inspect the topology for further details. WDYT?

pkg/prepare/prepare.go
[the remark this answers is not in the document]
dtomcej: Ah now I get it. Yes, you are correct, we can remove this in another issue.

Dockerfile
```diff
@@ -22,7 +22,7 @@ WORKDIR /go/src/github.com/containous/maesh
 RUN curl -sfL https://install.goreleaser.com/github.com/goreleaser/goreleaser.sh | sh
 
 # Download golangci-lint binary to bin folder in $GOPATH
-RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin v1.23.8
+RUN curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin v1.27.0
```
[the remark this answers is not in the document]
dtomcej: ikr

pkg/prepare/prepare.go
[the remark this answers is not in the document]
dtomcej: https://github.com/containous/maesh/issues/583

pkg/provider/provider.go
dtomcej: @jspdown was there a reason that this code didn't return an error and merely logged the error in the topology?

pkg/topology/builder.go
```diff
@@ -745,3 +779,16 @@ func mapCopy(m map[Key]struct{}) map[Key]struct{} {
 
 	return cpy
 }
+
+func removeTrafficSplitsLoop(loopDetected map[*Service][]Key) {
```
Ullaakut, line 783: Still confusing to me that we take a map called loopDetected that doesn't just contain booleans but information that is more complex. I'd suggest renaming this variable, but I don't know how. I'll LGTM though as not to block the PR, but I think this should be changed (or explained)

pkg/provider/provider.go
dtomcej: Will open a follow up issue for this.

dtomcej: https://github.com/containous/maesh/issues/585

pkg/topology/builder.go
dtomcej: https://github.com/containous/maesh/issues/586

PR #582, opened 2020-05-14
Fix incomplete topology when destination port is not found

### LINKED ISSUE #581: Incomplete Topology when a Service evaluation fails (sporadic)
### What did you do?

Let say you have created 2 Services and a TrafficTarget. In the TrafficTarget you have defined a destination with a port and a service-account which is used by pods matching the selector of both services. If one of the Services doesn't expose the port mentioned in the TrafficTarget you are going to receive an error:

```
Error building topology for TrafficTarget "tt@ns": unable to find destination ports on Service "svc-2@ns": destination port 80 of TrafficTarget "tt@ns" is not exposed by the service 
```

This error is expected. But if you modify the TrafficTarget specs multiple time and check at the dynamic config, sometimes, there's no router for the first service (the one that was supposed to work).

I expected to see a router for the first service. 

### What did you see instead?

There's sometimes no services.

### Output of controller log: (_What version of Maesh are you using?_)

Version: master

```
  --namespace=maesh \
  --set image.pullPolicy=IfNotPresent \
  --set image.name=maesh-dev \
  --set image.tag=latest \
  --set limits.tcp=20 \
  --set limits.http=20 \
  --set smi.deploy=true \
  --set acl=true
```

jspdown:
## What does this PR do?

This PR fixes a bug that happen when a TrafficTarget is applied on more than one Service and one of them is causing an error: If the `TrafficTarget.destination.port` is not present on the targeted service an error gets logged and it won't process next services.

Fixes #581 

In order to fix this behavior, instead of exiting the function on the first error, we just skip the Service.

## How to test it:

- Deploy 2 Service:
  - One which expose port 80
  - One which expose port 8080
- Define a TrafficTarget with a destination port = 80
- Update the TrafficTarget to trigger update (such as modifying the spec of a HTTPRouteGroup)

You should see one router for the traffic target in the dynamic config.

### DIFF
--- a/pkg/topology/builder.go
+++ b/pkg/topology/builder.go
@@ -111,7 +111,7 @@ func (b *Builder) evaluateTrafficTarget(res *resources, topology *Topology, tt *
 			setTrafficTargetWithErr(topology, tt, sources, err)
 			b.Logger.Errorf("Error building topology for TrafficTarget %q: %v", Key{tt.Name, tt.Namespace}, err)
 
-			return
+			continue
 		}
 
 		var destPods []Key
@@ -132,7 +132,7 @@ func (b *Builder) evaluateTrafficTarget(res *resources, topology *Topology, tt *
 			setTrafficTargetWithErr(topology, tt, sources, err)
 			b.Logger.Errorf("Error building topology for TrafficTarget %q: %v", Key{tt.Name, tt.Namespace}, err)
 
-			return
+			continue
 		}
 
 		// Create the ServiceTrafficTarget for the given service.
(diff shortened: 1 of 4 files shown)


pkg/topology/builder.go
```diff
@@ -218,9 +218,11 @@ func (b *Builder) evaluateTrafficSplit(topology *Topology, trafficSplit *split.T
...
 
-			return
+			continue
 		}
 
+		var err error
+
 		// As required by the SMI specification, backends must expose at least the same ports as the Service on
 		// which the TrafficSplit is.
 		for _, svcPort := range svc.Ports {
```
kevinpollet, line 228: It's a bit hard to understand these loops. Maybe splitting these loops will make the code easier to understand?

dtomcej: We refactor this in #550. Did we want to wait for that PR to merge and then update with this fix?

kevinpollet: You're right, I think it's simpler to merge #550 and then update this fix.

jspdown:
I thought this patch might deserve a unit test. While doing this I discovered an issue in `setTrafficSplitWithErr` and `setTrafficSplitWithErr`. They both missed a lot of property from the the TrafficTarget and TrafficSplit. I fixed this as well.

jspdown:
This PR will be blocked until https://github.com/containous/maesh/pull/564 gets merged.

Ullaakut:
@jspdown Do you need help with this PR? is there something else still blocking it?

jspdown:
@Ullaakut Thanks, that's fine. I just need to rebase it and fix the lint errors.

PR #584, opened 2020-05-14
Build kubernetes clients separately

### LINKED ISSUE #580: Warning message:  Neither --kubeconfig nor --master was specified
## Bug Report

While installing the latest version of Maesh I noticed there was a warning in the logs.

### What did you do?

Install Maesh using the Helm chart. No specific option.

No warning or errors

### What did you see instead?

```
W0514 10:10:30.195083       1 client_config.go:543] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
```

### Output of controller log: (_What version of Maesh are you using?_)

```
time="2020-05-14T10:10:30Z" level=info msg="Configuration loaded from flags."
time="2020-05-14T10:10:30Z" level=debug msg="Starting maesh prepare..."
time="2020-05-14T10:10:30Z" level=debug msg="Using masterURL: \"\""
time="2020-05-14T10:10:30Z" level=debug msg="Using kubeconfig: \"\""
W0514 10:10:30.195083       1 client_config.go:543] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
time="2020-05-14T10:10:30Z" level=debug msg="Building Kubernetes Client..."
time="2020-05-14T10:10:30Z" level=debug msg="Building SMI Access Client..."
time="2020-05-14T10:10:30Z" level=debug msg="Building SMI Specs Client..."
time="2020-05-14T10:10:30Z" level=debug msg="Building SMI Split Client..."
time="2020-05-14T10:10:30Z" level=info msg="Checking DNS provider"
time="2020-05-14T10:10:30Z" level=info msg="Checking CoreDNS"
time="2020-05-14T10:10:30Z" level=info msg="CoreDNS match"
time="2020-05-14T10:10:30Z" level=debug msg="ACL mode enabled: true"
time="2020-05-14T10:10:30Z" level=debug msg="Initializing API"
time="2020-05-14T10:10:30Z" level=debug msg="Initializing mesh controller"
time="2020-05-14T10:10:30Z" level=debug msg="Starting API"
time="2020-05-14T10:10:30Z" level=debug msg="Starting Informers"
time="2020-05-14T10:10:30Z" level=debug msg="API readiness enabled"
```

- Maesh: master

```
  --namespace=maesh \
  --set image.pullPolicy=IfNotPresent \
  --set image.name=maesh-dev \
  --set image.tag=latest \
  --set limits.tcp=20 \
  --set limits.http=20 \
  --set smi.deploy=true \
  --set acl=true
```

dtomcej:
## What does this PR do?

This PR:

- Adds checks to determine if an in-cluster or external-cluster config needs to be built
- Adds logging for what client is being built

Fixes #580

## Additional Notes

The warning comes from: https://github.com/kubernetes/client-go/blob/v0.17.4/tools/clientcmd/client_config.go#L543

Where an in-cluster config is built if the args are not present. We can do the check beforehand, and return the same client before calling this method, and skip that warning altogether.

### DIFF
--- a/pkg/k8s/client.go
+++ b/pkg/k8s/client.go
@@ -2,19 +2,20 @@ package k8s
 
 import (
 	"fmt"
+	"os"
 
 	accessclient "github.com/servicemeshinterface/smi-sdk-go/pkg/gen/client/access/clientset/versioned"
 	specsclient "github.com/servicemeshinterface/smi-sdk-go/pkg/gen/client/specs/clientset/versioned"
 	splitclient "github.com/servicemeshinterface/smi-sdk-go/pkg/gen/client/split/clientset/versioned"
 	"github.com/sirupsen/logrus"
-	kubeclient "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
 // Client is an interface for the various resource controllers.
 type Client interface {
-	GetKubernetesClient() kubeclient.Interface
+	GetKubernetesClient() kubernetes.Interface
 	GetAccessClient() accessclient.Interface
 	GetSpecsClient() specsclient.Interface
 	GetSplitClient() splitclient.Interface
@@ -25,7 +26,7 @@ var _ Client = (*ClientWrapper)(nil)
 
 // ClientWrapper holds the clients for the various resource controllers.
 type ClientWrapper struct {
-	kubeClient   *kubeclient.Clientset
+	kubeClient   *kubernetes.Clientset
 	accessClient *accessclient.Clientset
 	specsClient  *specsclient.Clientset
 	splitClient  *splitclient.Clientset
@@ -33,7 +34,7 @@ type ClientWrapper struct {
 
 // NewClient creates and returns a ClientWrapper that satisfies the Client interface.
 func NewClient(log logrus.FieldLogger, url string, kubeConfig string) (Client, error) {
-	config, err := clientcmd.BuildConfigFromFlags(url, kubeConfig)
+	config, err := buildConfig(log, url, kubeConfig)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


pkg/k8s/client.go
```diff
@@ -66,8 +67,24 @@ func NewClient(log logrus.FieldLogger, url string, kubeConfig string) (Client, e
 	}, nil
 }
 
+// buildConfig takes the url and kubeconfig, and returns an external or internal config.
+func buildConfig(log logrus.FieldLogger, url string, kubeConfig string) (*rest.Config, error) {
```
kevinpollet, line 71:
```suggestion
func buildConfig(log logrus.FieldLogger, masterURL, kubeConfig string) (*rest.Config, error) {
```

pkg/k8s/client.go
```diff
@@ -66,8 +67,24 @@ func NewClient(log logrus.FieldLogger, url string, kubeConfig string) (Client, e
 	}, nil
 }
 
+// buildConfig takes the url and kubeconfig, and returns an external or internal config.
```
kevinpollet, line 70:
```suggestion
// buildConfig takes the master URL and kubeconfig, and returns an external or internal config.
```

pkg/k8s/client.go
```diff
@@ -66,8 +67,24 @@ func NewClient(log logrus.FieldLogger, url string, kubeConfig string) (Client, e
...
+		log.Infoln("Creating in-cluster client")
+		return rest.InClusterConfig()
+	}
+
+	if url != "" && kubeConfig != "" {
+		log.Infoln("Creating cluster-external client from masterURL and kubeconfig")
+		return clientcmd.BuildConfigFromFlags(url, kubeConfig)
+	}
+
+	return nil, fmt.Errorf("could not create client: Missing masterURL or kubeConfig")
```
kevinpollet, line 83:
```suggestion
	return nil, fmt.Errorf("could not create client: missing masterURL or kubeConfig")
```

pkg/k8s/client.go
```diff
@@ -2,19 +2,20 @@ package k8s
...
-	kubeclient "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
 // Client is an interface for the various resource controllers.
 type Client interface {
-	GetKubernetesClient() kubeclient.Interface
+	GetKubernetesClient() kubernetes.Interface
```
kevinpollet, line 18: Maybe we could remove the `Get` prefix in client getters name?

As described [here](https://golang.org/doc/effective_go.html#Getters), the getter method name should be `KubernetesClient()`.

> There's nothing wrong with providing getters and setters yourself, and it's often appropriate to do so, but it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner.

pkg/k8s/client.go
kevinpollet: Same suggestion for the `NewClient` function.

pkg/k8s/client.go
```diff
@@ -66,31 +67,47 @@ func NewClient(log logrus.FieldLogger, url string, kubeConfig string) (Client, e
...
-func (w *ClientWrapper) GetKubernetesClient() kubeclient.Interface {
+// buildConfig takes the master URL and kubeconfig, and returns an external or internal config.
+func buildConfig(log logrus.FieldLogger, masterURL, kubeConfig string) (*rest.Config, error) {
+	if os.Getenv("KUBERNETES_SERVICE_HOST") != "" && os.Getenv("KUBERNETES_SERVICE_PORT") != "" {
+		// If these env vars are set, we can build an in-cluster config.
+		log.Infoln("Creating in-cluster client")
+		return rest.InClusterConfig()
+	}
+
+	if masterURL != "" && kubeConfig != "" {
```
kevinpollet, line 78: After thinking a bit more about it, I think that the `masterURL` is only used as a config [override](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/client_config.go#L569) and is not mandatory (e.g. I'm using only the `kubeconfig` flag when I'm running Maesh binary). So the test should look like my suggestion and error messages should be updated accordingly.

```suggestion
	if kubeConfig != "" {
```

WDYT?

dtomcej: Turns out I do the opposite! I only use MasterURL, since I use kubectl proxy :D

I should make this `&&` an `||`, and update that message.

PR #587, opened 2020-05-15
Remove Preparer interface

### LINKED ISSUE #583: Remove interface from Prepare package
### Proposal

The Prepare package does not need an interface since we are not mocking it or extending it.

dtomcej:
## What does this PR do?

This PR:

- Removes the Preparer interface

Fixes #583


dtomcej:
Naw, I like the prepare struct name as is.

PR #589, opened 2020-05-15
Patch and unpatch DNS configmap 

### LINKED ISSUE #576: CoreDNS config is not updated when Maesh is reinstalled in another namespace
### What did you do?

- Installed Maesh in the `default` ns: `helm install maesh maesh/maesh`.
- Uninstalled Maesh: `helm uninstall maesh`
- Reinstalled Maesh in the `maesh` ns: `helm install maesh maesh/maesh --namespace=maesh`
- Deployed the attached [services](https://github.com/containous/maesh/files/4622768/testcase.zip).
- Sent a request: `kubectl exec deployment/client -- curl whoami.default.maesh`

`HTTP 200 OK`

### What did you see instead?

`curl: (6) Could not resolve host: whoami.default.maesh`

dtomcej:
## What does this PR do?

This PR:

- Moves DNS methods to its own package
- Allows patching and unpatching to be done on DNS configmaps
- Allows patching functions to be idempotent

Fixes #576


pkg/prepare/prepare.go
```diff
@@ -175,6 +181,25 @@ func (p *Prepare) ConfigureCoreDNS(clusterDomain, maeshNamespace string) error {
...
+// backupConfigMap backs up a configmap with `-backup` appended to its name.
+func (p *Prepare) backupConfigMap(configMap *corev1.ConfigMap) error {
+	newConfigMap := configMap.DeepCopy()
+
+	newConfigMap.Name = configMap.Name + "-backup"
+
+	// Remove resourceVersion since it is not to be set manually.
+	newConfigMap.ObjectMeta.ResourceVersion = ""
+
+	if _, err := p.client.GetKubernetesClient().CoreV1().ConfigMaps(newConfigMap.Namespace).Get(newConfigMap.Name, metav1.GetOptions{}); err == nil {
```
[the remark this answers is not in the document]
dtomcej: >Could we update the ConfigMap if it already exists?

Sure, but I'm wondering what the use case is where there is already an existing backup. If there is already a backup, then we may risk overwriting a pre-patched backup with one that is patched...

If we don't restore properly, then the modified configmap will be backed up, causing more issues.

>We can't be sure there is no `configMap.Name + "-backup" already present in the namespace. I suggest we put this backup in Maesh namespace

Ya, that makes sense.

jspdown:
> Do we want to restore in the cleanup step?

Yes it does make sense. We should be able to install + cleanup + install and get the same behavior.

> Do we want to reuse the same getCoreDNSConfigmap etc logic from the prepare package to handle this so that we don't duplicate code?

Yes, it would be better to take the configuration from the core-dns deployment than directly from the configmap. At least we can be sure that's the one being used.

> How do we want to handle failure on backup or restore?

I'm not sure we can do something. If we can't find or create the configmap it means either the resource doesn't exist, we can't reach the cluster or there's a bug on our side.

So, I would say, just display an error so the user can see it.

dtomcej:
Still needs cleanup tests

pkg/prepare/prepare.go
```diff
@@ -168,13 +177,49 @@ func (p *Prepare) ConfigureCoreDNS(clusterDomain, maeshNamespace string) error {
...
+	if _, err := p.client.KubernetesClient().CoreV1().ConfigMaps(newConfigMap.Namespace).Create(newConfigMap); err != nil {
+		return nil, err
+	}
+
+	// Label the configmap as backed up.
+	if configMap.ObjectMeta.Labels == nil {
+		configMap.ObjectMeta.Labels = make(map[string]string)
+	}
+
+	configMap.ObjectMeta.Labels["maesh-backed-up"] = "true"
```
kevinpollet, line 217: I feel a little anxious every time I see this kind of magic string used everywhere in the code. 
Why not create a constant?

kevinpollet: Same comment the `maesh-patched` string.

pkg/cleanup/cleanup.go
```diff
@@ -11,14 +16,18 @@ type Cleanup struct {
 	client    k8s.Client
 	log       logrus.FieldLogger
 	namespace string
+	prep      *prepare.Prepare
```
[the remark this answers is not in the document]
dtomcej: Moving to a DNS package would allow us to not deal with backups, and only add/remove as needed.

I will use that as a basis to move this issue forward.

integration/try/try.go
```diff
@@ -182,6 +182,8 @@ func (t *Try) WaitCommandExecute(command string, argSlice []string, expected str
 			return fmt.Errorf("output %s does not contain %s", string(output), expected)
 		}
 
+		fmt.Print(string(output))
```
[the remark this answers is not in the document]
dtomcej: Yes, it allows us to log the helm command responses, which are lost otherwise.

pkg/dns/dns.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+
+	c.log.Info("CoreDNS match")
+
+	return true, nil
+}
+
+func isCoreDNSVersionSupported(versionLine string) bool {
+	for _, v := range supportedCoreDNSVersions {
+		if strings.Contains(versionLine, v) || strings.Contains(versionLine, "v"+v) {
```
kevinpollet, line 117: This condition is not needed?
In both cases the string contains `v`.

```suggestion
		if strings.Contains(versionLine, v) {
```

pkg/dns/dns.go
```diff
@@ -0,0 +1,452 @@
...
+	c.log.Info("Checking KubeDNS")
+
+	_, err := c.client.KubernetesClient().AppsV1().Deployments(metav1.NamespaceSystem).Get("kube-dns", metav1.GetOptions{})
+	if kubeerror.IsNotFound(err) {
+		c.log.Debugf("KubeDNS does not exist in namespace %q", metav1.NamespaceSystem)
+		return false, nil
+	}
+
+	if err != nil {
+		return false, fmt.Errorf("unable to get deployment %q in namesapce %q: %w", "kube-dns", metav1.NamespaceSystem, err)
```
kevinpollet, line 135:
```suggestion
		return false, fmt.Errorf("unable to get KubeDNS deployment in namespace %q: %w", metav1.NamespaceSystem, err)
```

pkg/dns/dns.go
```diff
@@ -0,0 +1,452 @@
...
+	}
+
+	var (
+		serviceIP string
+		ebo       = backoff.NewConstantBackOff(10 * time.Second)
+	)
+
+	c.log.Debug("Getting CoreDNS service IP")
+
+	if err = backoff.Retry(safe.OperationWithRecover(func() error {
```
kevinpollet, line 259: Thoughts. If I correctly understand the code we are waiting for the CoreDNS service to be available (service installed by our Helm chart if `kubedns` option is true). 

Maybe we can remove the duplicated CoreDNS config for Maesh and apply a patch here? or create a `configMap` with our config that could be used by the CoreDNS deployment? Does it make sense?

dtomcej: Not sure if I understand.

This IP is the service IP of our coreDNS configuration, that we need to have added as a stubDomain in kubeDNS.

We need this IP to have been assigned before we can successfully patch the kubeDNS configmap to forward requests.

kevinpollet: I'm just saying that we could reuse the logic to patch our CoreDNS deployment which is added as a stubDomain in kubeDNS. This will avoid us to have a duplicate CoreDNS config, one in the code and one in the Helm chart.

Does it make sense?

dtomcej: Ohhh, so we deploy ours with an "empty" config...and then patch both our coreDNS and the existing kubeDNS?

hmmmm yep, I like that idea

dtomcej: https://github.com/containous/maesh/issues/593

pkg/dns/dns.go
```diff
@@ -0,0 +1,453 @@
...
+	match, err = c.kubeDNSMatch()
+	if err != nil {
+		return UnknownDNS, err
+	}
+
+	if match {
+		return KubeDNS, nil
+	}
+
+	return UnknownDNS, fmt.Errorf("no supported DNS service available for installing maesh: %w", err)
```
kevinpollet, line 80: For me, it's a bit strange to always return `UnknownDNS` with a corresponding error. 

Is the `UnknownDNS` value really needed? 
Why return an error here if there is no error?

dtomcej: Well, the error is that none of the supported DNS services are available, which means Maesh will not work.

If we returned `*Provider` instead, we could return nil instead of `UnknownDNS`. Would that be better?

kevinpollet: I see, returning `Provider` is fine.

PR #590, opened 2020-05-19
Add forwardingTimeouts

### LINKED ISSUE #577: Forwarding Timeouts should be configurable
### Proposal

Forwarding timouts should be configurable for Maesh nodes.

### Background

Default timeouts may not work for all users.

dtomcej:
## What does this PR do?

This PR:

- Adds forwardingTimeouts to the helm chart

Fixes #577

### DIFF
--- a/helm/chart/maesh/values.yaml
+++ b/helm/chart/maesh/values.yaml
@@ -41,10 +41,12 @@ mesh:
   # Added so we can launch on nodes with restrictions
   nodeSelector: {}
   tolerations: []
-  forwardingTimeouts:
-    dialTimeout: "30s"
-    responseHeaderTimeout: "0s"
-    idleConnTimeout: "1s"
+  # forwardingTimeouts:
+  #   dialTimeout: "30s"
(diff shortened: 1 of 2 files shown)


helm/chart/maesh/values.yaml
```diff
@@ -41,7 +41,10 @@ mesh:
   # Added so we can launch on nodes with restrictions
   nodeSelector: {}
   tolerations: []
-
+  forwardingTimeouts:
```
kevinpollet, line 44: I think that by default we should use the default values from Træfik, wdyt?

These options should be optional and commented as described [here](https://github.com/containous/maesh/blob/master/helm/chart/maesh/Guidelines.md). Something like:

```
# forwardingTimeouts:
#    dialTimeout: "30s"
#    responseHeaderTimeout: "0s"
#    idleConnTimeout: "1s"
```

dtomcej: Good Catch!

PR #592, opened 2020-05-21
Rework controller loop to use a work queue

### LINKED ISSUE #514: Retrieve port mapping from existing shadow services

### LINKED ISSUE #562: Shadow service manager should be idempotent
### Proposal

To forward traffic through `mesh` proxies, Maesh uses shadow services. To create, update or delete a shadow service corresponding to a user `service` the Maesh controller subscribes to `service` events emitted by informers.

However, it could be possible that the controller misses some events which can lead to bad state and behaviour. e.g. if the controller is down and a user service is deleted, on the next deployment, the shadow service is not [created or updated](https://github.com/containous/maesh/blob/master/pkg/controller/service.go#L54) which will not reflect potential changes.

To make sure that the shadow services state reflects the last known cluster state, the shadow service manager could be idempotent. Also, a service update event could create the corresponding shadow service if it does not exist to deal with missing create events.

kevinpollet:
## What does this PR do?

This PR:

- Removes unused code.
- Renames some test packages for consistency.
- Reworks the controller loop to use a work queue.
- Reworks the `ShadowServiceManager` to make it `idempotent`.

Fixes #562.

## Additional Notes

To fix #514, the TCP tables must be built from the `informers` cache which requires some synchronisation before starting the events processing. As discussed, to prepare the work on #514, I've reworked the controller loop to use a work queue which allows enqueueing before processing events.

By design, work queue keys are unique. So, if we receive a lot of events with the same work key, only one will be present in the queue at a time. This would avoid us some unnecessary work if we received a lot of events which require to rebuild the configuration, for example.

Something to keep in mind is that we are building the desired state of the configuration and the shadow services from the last know state of the cluster, the informers `cache`. The controller work is to update the configuration and the shadow services to match this state. So, if a service is updated `10000` times and is deleted, it's not necessary to process all events, our job is to remove the shadow service.

As we should not rely on an expected cluster state, I've also modified the `ShadowServiceManager` to make it `idempotent`. I'm pretty sure that it could be simplified and I will try to address that in my following work on #514.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/containous/maesh/pkg/annotations"
@@ -198,7 +199,7 @@ func (c *Controller) init() {
 }
 
 // Run is the main entrypoint for the controller.
-func (c *Controller) Run(stopCh <-chan struct{}) error {
+func (c *Controller) Run(stopCh <-chan struct{}) {
 	// Handle a panic with logging and exiting.
 	defer utilruntime.HandleCrash()
 
@@ -221,8 +222,6 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 
 	<-stopCh
 	c.logger.Info("Shutting down workers")
-
-	return nil
 }
 
 // startInformers starts the controller informers.
@@ -305,7 +304,7 @@ func (c *Controller) processNextWorkItem() bool {
 
 	if key != configRefreshKey {
 		if err := c.syncShadowService(key.(string)); err != nil {
-			c.handleErr(key, err)
+			c.handleErr(key, fmt.Errorf("unable to sync shadow service: %w", err))
 			return true
 		}
 	}
@@ -313,7 +312,7 @@ func (c *Controller) processNextWorkItem() bool {
 	// Build and store config.
 	topo, err := c.topologyBuilder.Build(c.ignoredResources)
 	if err != nil {
-		c.handleErr(key, err)
+		c.handleErr(key, fmt.Errorf("unable to build topology: %w", err))
 		return true
 	}
 
@@ -356,6 +355,6 @@ func (c *Controller) handleErr(key interface{}, err error) {
 		return
 	}
 
-	c.logger.Errorf("Unable to complete work for key %q: %v", key, err)
+	c.logger.Errorf("Unable to complete work %q: %v", key, err)
 	c.workQueue.Forget(key)
 }
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -2,7 +2,6 @@ package controller
 
 import (
 	"context"
-	"reflect"
 	"time"
 
 	"github.com/containous/maesh/pkg/annotations"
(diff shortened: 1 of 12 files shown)


pkg/controller/service.go
```diff
@@ -82,119 +80,104 @@ func (s *ShadowServiceManager) Create(userSvc *corev1.Service) error {
...
-		return err
-	}
+	for _, shadowSvcPort := range shadowSvc.Spec.Ports {
+		if containsPort(svc.Spec.Ports, shadowSvcPort) {
+			continue
+		}
 
-	if err := s.kubeClient.CoreV1().Services(s.namespace).Delete(name, &metav1.DeleteOptions{}); err != nil {
-		return err
+		s.removeServicePortMapping(svc.Namespace, svc.Name, shadowSvcPort)
```
jspdown, line 149: I think there's a problem here.
If the `shadowSvcPort` is part of the svc ports, the traffic-type of the service can still be different. In this case we should call `s.removeServicePortMapping` to make sure we are not leaking ports.
If you have port 8080 reserved in the TCP state table and the service become UDP, we have to remove this port from the TCP state table.

jspdown: Ah my bad I didn't see the `cleanupPortMappings`. I find this function a bit disturbing, we are cleaning up ports according to the old traffic-type if the port has changed but not if it didn't. So we are relying on the call to `cleanupPortMapping` to cleanup what we left.

pkg/controller/controller.go
```diff
@@ -209,27 +216,13 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
...
-
-			conf := c.provider.BuildConfig(topo)
-
-			if !reflect.DeepEqual(c.currentConfiguration.Get(), conf) {
-				c.currentConfiguration.Set(conf)
-			}
-		}
-	}
+	// Start to poll work from the queue.
+	go wait.Until(c.runWorker, time.Second, stopCh)
```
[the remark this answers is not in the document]
kevinpollet: It's not possible, the `runWorker` call blocks until a message is available in the queue as explained in `Get` documentation of the work queue:

>// Get blocks until it can return an item to be processed. If shutdown = true,
// the caller should end their goroutine. You must call Done with item when you
// have finished processing it.

pkg/controller/service.go
kevinpollet: Agreed but I haven't found a better solution yet. 

Thoughts?

dtomcej: What if we tied port mappings to the `service.UID`? This value is constant and does not change on object update.

That way we could easily keep track of ports regardless of the changes made to the user service.

kevinpollet: I'm not sure to understand? 
The problem here is that before allocation we have to remove

dtomcej: Sorry, I misunderstood.

PR #608, opened 2020-05-27
Move API outside of the Controller

### LINKED ISSUE #494: Provider Interface
#### Proposal

Currently Maesh has its own provider interface, but we should probably follow the Traefik Provider interface to allow better integration when using Traefik and Maesh in the same project.

jspdown:
## What does this PR do?

This PR moves the API outside of the Controller and exposes the Topology in the API on the `/api/topology/current`.

Other changes:
- Graceful shutdown of the API
- Removed the `signals` package in favor of Traefik's `cmd.ContextWithSignal` function.
- Changed the output of the API in case of an InternalError. If there's an internal error, the operator should be the only one to see what's the problem by looking at the logs. Users should just be notified that something went wrong.
- Removed the handler field of the Controller as it's not needed

Fixes #494

### How to test it

* `make test`
* `make test-integration`


cmd/maesh/maesh.go
```diff
@@ -106,32 +111,72 @@ func maeshCommand(iConfig *cmd.MaeshConfiguration) error {
...
 		MaxUDPPort:       minUDPPort + iConfig.LimitUDPPort,
-	}, log)
+	}, apiServer, log)
 	if err != nil {
 		return fmt.Errorf("unable to create controller: %w", err)
 	}
 
-	// run the ctr loop to process items
-	ctr.Run(stopCh)
+	errCh := make(chan error, 2)
```
kevinpollet, line 137:
```suggestion
	errCh := make(chan error)
```

cmd/maesh/maesh.go
```diff
@@ -106,32 +111,72 @@ func maeshCommand(iConfig *cmd.MaeshConfiguration) error {
...
+
+	var wg sync.WaitGroup
```
kevinpollet, line 139: Thought, is it really needed to run the `api` and the `controller` in a go routine?

pkg/api/api.go
```diff
@@ -43,97 +43,138 @@ type podInfo struct {
...
 
-// Init handles any api initialization.
-func (a *API) Init() error {
-	a.log.Debugln("Initializing API")
+// SetReadiness sets the readiness flag in the API.
+func (a *API) SetReadiness(isReady bool) {
+	readiness, ok := a.readiness.Get().(bool)
 
-	a.router = mux.NewRouter()
+	if ok && readiness == isReady {
```
kevinpollet, line 103: I don't think that this test is needed.

pkg/api/api.go
```diff
@@ -43,97 +43,138 @@ type podInfo struct {
...
+	}
 
-	a.router.HandleFunc("/api/configuration/current", a.getCurrentConfiguration)
-	a.router.HandleFunc("/api/status/nodes", a.getMeshNodes)
-	a.router.HandleFunc("/api/status/node/{node}/configuration", a.getMeshNodeConfiguration)
-	a.router.HandleFunc("/api/status/readiness", a.getReadiness)
+	a.readiness.Set(isReady)
 
-	return nil
+	if isReady {
```
kevinpollet, line 109: I don't think that this test is needed, maybe it's enough to display the ready value in the debug message?

pkg/api/api.go
```diff
@@ -43,97 +43,138 @@ type podInfo struct {
...
 }
 
 // getMeshNodes returns a list of mesh nodes visible from the controller, and some basic readiness info.
-func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
-	podInfoList := []podInfo{}
-
-	sel := labels.Everything()
+func (a *API) getMeshNodes(w http.ResponseWriter, _ *http.Request) {
+	// Make sure it returns an empty array and not "null".
+	podInfoList := make([]podInfo, 0)
```
kevinpollet, line 171: Should be initialized after the pods retrieval.

```suggestion
	podInfoList := make([]podInfo, len(podList))
```

pkg/controller/controller.go
```diff
@@ -46,6 +45,13 @@ type PortMapper interface {
 	Remove(svc k8s.ServiceWithPort) (int32, error)
 }
 
+// SharedStore is capable sharing a state of the controller.
```
kevinpollet, line 48:
```suggestion
// SharedStore is used to share the controller state.
```

pkg/controller/controller.go
```diff
@@ -199,7 +200,18 @@ func (c *Controller) init() {
 }
 
 // Run is the main entrypoint for the controller.
-func (c *Controller) Run(stopCh <-chan struct{}) {
+func (c *Controller) Run(ctx context.Context) {
+	stopCh := make(chan struct{})
```
kevinpollet, line 204: Is it possible to keep the `stopCh` as a param?

pkg/api/api.go
```diff
@@ -43,97 +43,138 @@ type podInfo struct {
...
+// getCurrentConfiguration returns the current configuration.
+func (a *API) getCurrentConfiguration(w http.ResponseWriter, _ *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	if err := json.NewEncoder(w).Encode(a.configuration.Get()); err != nil {
+		a.log.Errorf("Unable to serialize dynamic configuration: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
 
-		a.readiness = true
+		return
```
dtomcej, line 134: is this `return` needed?

pkg/api/api.go
```diff
@@ -43,97 +43,138 @@ type podInfo struct {
...
+func (a *API) getCurrentTopology(w http.ResponseWriter, _ *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 
-	if err := json.NewEncoder(w).Encode(a.currentConfiguration.Get()); err != nil {
-		a.log.Error(err)
+	if err := json.NewEncoder(w).Encode(a.topology.Get()); err != nil {
+		a.log.Errorf("Unable to serialize topology: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
+
+		return
```
dtomcej, line 146: is this `return` needed?

pkg/api/api.go
```diff
@@ -43,97 +43,138 @@ type podInfo struct {
...
 
 	w.Header().Set("Content-Type", "application/json")
 
-	if err := json.NewEncoder(w).Encode(a.readiness); err != nil {
-		a.log.Error(err)
+	if err := json.NewEncoder(w).Encode(isReady); err != nil {
+		a.log.Errorf("Unable to serialize readiness: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
+
+		return
```
dtomcej, line 164: is this `return` needed?

pkg/api/api.go
```diff
@@ -159,7 +200,10 @@ func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 
 	if err := json.NewEncoder(w).Encode(podInfoList); err != nil {
-		a.log.Error(err)
+		a.log.Errorf("Unable to serialize mesh nodes: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
+
+		return
```
dtomcej, line 206: is this `return` needed?

pkg/api/api.go
```diff
@@ -170,43 +214,57 @@ func (a *API) getMeshNodeConfiguration(w http.ResponseWriter, r *http.Request) {
...
 	}
 
 	w.Header().Set("Content-Type", "application/json")
 
 	if _, err := w.Write(body); err != nil {
-		a.log.Error(err)
+		a.log.Errorf("Unable to write mesh nodes: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
+
+		return
```
dtomcej, line 254: is this `return` needed?

pkg/controller/controller.go
```diff
@@ -98,7 +101,7 @@ type Controller struct {
 
 // NewMeshController builds the informers and other required components of the mesh controller, and returns an
 // initialized mesh controller object.
-func NewMeshController(clients k8s.Client, cfg Config, logger logrus.FieldLogger) (*Controller, error) {
+func NewMeshController(clients k8s.Client, cfg Config, store SharedStore, logger logrus.FieldLogger) (*Controller, error) {
```
dtomcej, line 104: Should this return the stopChan as well as the Controller? so that we can use the stop chan to stop the controller, instead of having to use the context to do so?

pkg/controller/controller.go
dtomcej: Could we keep the `stopCh` as a controller member, and return it when the controller is created, then we can just use it to stop here instead of requiring a context?

cmd/maesh/maesh.go
```diff
@@ -106,32 +111,72 @@ func maeshCommand(iConfig *cmd.MaeshConfiguration) error {
...
+	// Wait for a stop event.
+	select {
+	case <-ctx.Done():
+	case err := <-errCh:
+		log.Error(err)
+	}
+
+	// Shutdown gracefully the API and the Controller.
+	go ctr.Stop()
+	go func() {
```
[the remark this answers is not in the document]
jspdown: Definitely not, you're right!

cmd/maesh/maesh.go
jspdown: I think it is. At least one of them must be run in a goroutine as `api.ListenAndServe` and `ctrl.Start` are blocking. Then if we don't run both in a goroutine, if the API stops with an error we won't to be able to stop the controller.
So the two options I see are:
- Run the `ListenAndServe` and `Start` in 2 goroutine.
- Or just run the stop mechanism in a goroutine.

pkg/api/api.go
jspdown: The goal was to avoid printing the debug log twice if the value hasn't changed. I agree, that's superfluous. I'm going to simplify that.

pkg/controller/controller.go
jspdown: I found an alternative solution, removing the need of a `Stop` method and the need of a taking a context in the `Run` method.

pkg/controller/controller.go
```diff
@@ -126,6 +129,8 @@ func NewMeshController(clients k8s.Client, cfg Config, logger logrus.FieldLogger
 		ignoredResources: ignoredResources,
 		tcpStateTable:    tcpStateTable,
 		udpStateTable:    udpStateTable,
+		store:            store,
+		stop:             make(chan struct{}),
```
dtomcej, line 133: Is this required anymore?

pkg/api/api.go
```diff
@@ -43,101 +43,124 @@ type podInfo struct {
...
-	if !a.readiness {
-		a.log.Debug("API readiness enabled")
+// getCurrentConfiguration returns the current configuration.
+func (a *API) getCurrentConfiguration(w http.ResponseWriter, _ *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
 
-		a.readiness = true
+	if err := json.NewEncoder(w).Encode(a.configuration.Get()); err != nil {
+		a.log.Errorf("Unable to serialize dynamic configuration: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
```
kevinpollet, line 121: Why not using https://golang.org/pkg/net/http/#Error?
```suggestion
		http.Error(w, "", http.StatusInternalServerError)
```

pkg/api/api.go
```diff
@@ -43,101 +43,124 @@ type podInfo struct {
...
-func (a *API) getCurrentConfiguration(w http.ResponseWriter, r *http.Request) {
+// getCurrentTopology returns the current topology.
+func (a *API) getCurrentTopology(w http.ResponseWriter, _ *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 
-	if err := json.NewEncoder(w).Encode(a.currentConfiguration.Get()); err != nil {
-		a.log.Error(err)
+	if err := json.NewEncoder(w).Encode(a.topology.Get()); err != nil {
+		a.log.Errorf("Unable to serialize topology: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
```
kevinpollet, line 131: See my previous comment.

pkg/api/api.go
```diff
@@ -43,101 +43,124 @@ type podInfo struct {
...
 }
 
 // getReadiness returns the current readiness value, and sets the status code to 500 if not ready.
-func (a *API) getReadiness(w http.ResponseWriter, r *http.Request) {
-	if !a.readiness {
-		w.WriteHeader(http.StatusInternalServerError)
+func (a *API) getReadiness(w http.ResponseWriter, _ *http.Request) {
+	isReady, _ := a.readiness.Get().(bool)
+	if !isReady {
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
```
kevinpollet, line 139: See my previous comment.

pkg/api/api.go
```diff
@@ -43,101 +43,124 @@ type podInfo struct {
...
+		return
 	}
 
 	w.Header().Set("Content-Type", "application/json")
 
-	if err := json.NewEncoder(w).Encode(a.readiness); err != nil {
-		a.log.Error(err)
+	if err := json.NewEncoder(w).Encode(isReady); err != nil {
+		a.log.Errorf("Unable to serialize readiness: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
```
kevinpollet, line 147: See my previous comment.

pkg/api/api.go
```diff
@@ -43,101 +43,124 @@ type podInfo struct {
...
-	requirement, err := labels.NewRequirement("component", selection.Equals, []string{"maesh-mesh"})
+func (a *API) getMeshNodes(w http.ResponseWriter, _ *http.Request) {
+	podList, err := a.podLister.List(labels.Everything())
 	if err != nil {
-		a.log.Error(err)
-	}
-
-	sel = sel.Add(*requirement)
+		a.log.Errorf("Unable to retrieve pod list: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
```
kevinpollet, line 156: See my previous comment.

pkg/api/api.go
```diff
@@ -148,65 +171,77 @@ func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
...
 		}
-		podInfoList = append(podInfoList, p)
 	}
 
 	w.Header().Set("Content-Type", "application/json")
 
 	if err := json.NewEncoder(w).Encode(podInfoList); err != nil {
-		a.log.Error(err)
+		a.log.Errorf("Unable to serialize mesh nodes: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
```
kevinpollet, line 185: Same here.

pkg/api/api.go
```diff
@@ -148,65 +171,77 @@ func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
...
 
 	if _, err := w.Write(body); err != nil {
-		a.log.Error(err)
+		a.log.Errorf("Unable to write mesh nodes: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
 	}
 }
 
-func (a *API) writeErrorResponse(w http.ResponseWriter, errorMessage string, status int) {
+func (a *API) writeErrorResponse(w http.ResponseWriter, err error, status int) {
```
kevinpollet, line 235: I don't think that this method is necessary.

pkg/api/api.go
```diff
@@ -148,65 +171,77 @@ func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
...
+		a.log.Errorf("Unable to get configuration from pod: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusBadGateway)
+
 		return
 	}
 
-	defer resp.Body.Close()
+	defer func() {
+		if closeErr := resp.Body.Close(); closeErr != nil {
+			a.log.Errorf("Unable to close response body: %w", closeErr)
```
kevinpollet, line 215:
```suggestion
			a.log.Errorf("Unable to close response body: %v, closeErr)
```

pkg/api/api.go
```diff
@@ -148,65 +171,77 @@ func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
...
 // getMeshNodeConfiguration returns the configuration for a named pod.
 func (a *API) getMeshNodeConfiguration(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 
-	pod, err := a.podLister.Pods(a.meshNamespace).Get(vars["node"])
+	pod, err := a.podLister.Pods(a.namespace).Get(vars["node"])
 	if err != nil {
 		if kubeerror.IsNotFound(err) {
-			a.writeErrorResponse(w, fmt.Sprintf("unable to find pod: %s", vars["node"]), http.StatusNotFound)
+			a.writeErrorResponse(w, fmt.Errorf("unable to find pod: %s", vars["node"]), http.StatusNotFound)
```
kevinpollet, line 196: I think we can use a string message here and the `http.Error` function.

pkg/api/api.go
```diff
@@ -148,65 +171,77 @@ func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
...
 			return
 		}
 
-		a.writeErrorResponse(w, fmt.Sprintf("unable to retrieve pod: %v", err), http.StatusInternalServerError)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
```
kevinpollet, line 200: Same here.

pkg/api/api.go
```diff
@@ -148,65 +171,77 @@ func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
...
 
 		return
 	}
 
 	resp, err := http.Get(fmt.Sprintf("http://%s:8080/api/rawdata", pod.Status.PodIP))
 	if err != nil {
-		a.writeErrorResponse(w, fmt.Sprintf("unable to get configuration from pod: %v", err), http.StatusBadGateway)
+		a.log.Errorf("Unable to get configuration from pod: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusBadGateway)
```
kevinpollet, line 208: Same here.

pkg/api/api.go
```diff
@@ -148,65 +171,77 @@ func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
...
+		if closeErr := resp.Body.Close(); closeErr != nil {
+			a.log.Errorf("Unable to close response body: %w", closeErr)
+		}
+	}()
 
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		a.writeErrorResponse(w, fmt.Sprintf("unable to get configuration response body from pod: %v", err), http.StatusBadGateway)
+		a.log.Errorf("Unable to get configuration response body from pod: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusBadGateway)
```
kevinpollet, line 222: Same here.

pkg/api/api.go
```diff
@@ -148,65 +171,77 @@ func (a *API) getMeshNodes(w http.ResponseWriter, r *http.Request) {
...
+
 		return
 	}
 
 	w.Header().Set("Content-Type", "application/json")
 
 	if _, err := w.Write(body); err != nil {
-		a.log.Error(err)
+		a.log.Errorf("Unable to write mesh nodes: %v", err)
+		a.writeErrorResponse(w, nil, http.StatusInternalServerError)
```
kevinpollet, line 231: Same here.

kevinpollet:
This PR will fix #494, right?

pkg/api/api.go
jspdown: I just followed what was already present in the code ;) Yeah It's definitely more appropriate.

cmd/maesh/maesh.go
```diff
@@ -92,34 +98,81 @@ func maeshCommand(iConfig *cmd.MaeshConfiguration) error {
...
+		if err := ctr.Run(ctrlStopCh); err != nil {
+			ctrlErrCh <- fmt.Errorf("controller has stopped unexpectedly: %w", err)
+		}
+	}()
+
+	// Wait for a stop event and shutdown servers.
+	select {
+	case <-ctx.Done():
+		ctrlStopCh <- struct{}{}
+
```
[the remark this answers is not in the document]
jspdown: Our linter won't allow this:
> only cuddled expressions if assigning variable or using from line above (wsl)

At some point, I would like to re-open the debate around the need of the wsl module.

PR #613, opened 2020-05-28
Add Helm keys to configure annotations for mesh deployment and pods

### LINKED ISSUE #612: Support custom deployment annotations in Helm chart
### What did you do?

Tried using maesh helm chart. There is no way to add deployment annotations via values.
I would like to add "sidecar.jaegertracing.io/inject": "true" to deployment annotations.

maesh helm chart should support custom deployment annotations via helm values.

dtomcej:
## What does this PR do?

This PR:

- Adds configurable annotations for Mesh deployment, and mesh pods

Fixes #612

### DIFF
--- a/helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
+++ b/helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
@@ -9,10 +9,10 @@ metadata:
     chart: {{ include "maesh.chartLabel" . | quote }}
     release: {{ .Release.Name | quote }}
     heritage: {{ .Release.Service | quote }}
-    {{- with .Values.mesh.annotations }}
+  {{- with .Values.mesh.annotations }}
   annotations:
 {{- toYaml . | nindent 4 }}
-    {{- end }}
+  {{- end }}
 spec:
   selector:
     matchLabels:
--- a/helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
+++ b/helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
@@ -9,6 +9,10 @@ metadata:
     chart: {{ include "maesh.chartLabel" . | quote }}
     release: {{ .Release.Name | quote }}
     heritage: {{ .Release.Service | quote }}
+    {{- with .Values.mesh.annotations }}
+  annotations:
+{{- toYaml . | nindent 4 }}
+    {{- end }}
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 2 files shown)


dtomcej:
This PR is implemented based on the pattern in https://github.com/containous/traefik-helm-chart/blob/master/traefik/templates/deployment.yaml

helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
```diff
@@ -9,6 +9,10 @@ metadata:
     chart: {{ include "maesh.chartLabel" . | quote }}
     release: {{ .Release.Name | quote }}
     heritage: {{ .Release.Service | quote }}
+    {{- with .Values.mesh.annotations }}
```
kevinpollet, line 12:
```suggestion
  {{- with .Values.mesh.annotations }}
```

helm/chart/maesh/templates/mesh/mesh-daemonset.yaml
```diff
@@ -9,6 +9,10 @@ metadata:
...
+  annotations:
+{{- toYaml . | nindent 4 }}
+    {{- end }}
```
kevinpollet, line 15:
```suggestion
  {{- end }}
```

helm/chart/maesh/values.yaml
```diff
@@ -65,6 +65,10 @@ mesh:
   # Added so we can launch on nodes with restrictions
   nodeSelector: {}
   tolerations: []
+  # Additional deployment annotations (e.g. for jaeger-operator sidecar injection)
+  annotations: {}
+  # Additional pod annotations (e.g. for mesh injection or prometheus scraping)
```
kevinpollet, line 70: I'm not sure that these examples are relevant because, we already have `prometheus scraping` annotations and injecting a mesh in a mesh does not make sense, right?

helm/chart/maesh/values.yaml
```diff
@@ -65,6 +65,10 @@ mesh:
   # Added so we can launch on nodes with restrictions
   nodeSelector: {}
   tolerations: []
+  # Additional deployment annotations (e.g. for jaeger-operator sidecar injection)
+  annotations: {}
```
kevinpollet, line 69: Just a question, the `jaeger-operator` sidecar injection is only required if our jaeger intergation/deployment is not used, right?

kevinpollet: If we follow the Helm chart [Guidelines](https://github.com/containous/maesh/blob/master/helm/chart/maesh/Guidelines.md) these options should be commented out. Does it make sense? wdyt?

PR #614, opened 2020-05-28
Move the defaultMode Helm key to be global

### LINKED ISSUE #605: Rename Helm option mesh.defaultMode to defaultMode
### Proposal

The Helm option `mesh.defaultMode` is currently used to define the default traffic type mode. 

As the default traffic type mode is not tied to `mesh` proxies and is more a `global` option like the `acl` one, it should be renamed to `defaultMode` and the `mesh.defaultMode` option should be deprecated.

dtomcej:
## What does this PR do?

This PR:

- Moves the defaultMode configuration to be global

Fixes #605

## Additional Notes

We may want to introduce a `global` key similar to Traefik's config, to handle these values, as they seem to be piling up.

### DIFF
--- a/docs/content/migration/helm-chart.md
+++ b/docs/content/migration/helm-chart.md
@@ -21,7 +21,9 @@ and `mesh.logLevel` options to configure the logging level for the controller an
 The `smi.enable` option has been deprecated and removed. You should use the new and backward compatible ACL mode 
 option as described in the [documentation](../install.md#access-control-list). 
 
+## v2.0 to v2.1
+
 ### Default Mode
 
-The `controller.mesh.defaultMode` option has been removed. You should use the new `defaultMode` option to configure
- the default traffic mode for Maesh services.
+The `controller.mesh.defaultMode` option has been deprecated and will be removed in a future major release.
+You should use the new `defaultMode` option to configure the default traffic mode for Maesh services.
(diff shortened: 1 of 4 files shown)


docs/content/migration/helm-chart.md
```diff
@@ -20,3 +20,8 @@ and `mesh.logLevel` options to configure the logging level for the controller an
 
 The `smi.enable` option has been deprecated and removed. You should use the new and backward compatible ACL mode 
 option as described in the [documentation](../install.md#access-control-list). 
+
+### Default Mode
+
+The `controller.mesh.defaultMode` option has been removed. You should use the new `defaultMode` option to configure
```
kevinpollet, line 26:
```suggestion
The `controller.mesh.defaultMode` option has been deprecated and will be removed in a futur Major release. You should use the new `defaultMode` option to configure
```

docs/content/migration/helm-chart.md
```diff
@@ -20,3 +20,8 @@ and `mesh.logLevel` options to configure the logging level for the controller an
 
 The `smi.enable` option has been deprecated and removed. You should use the new and backward compatible ACL mode 
 option as described in the [documentation](../install.md#access-control-list). 
+
+### Default Mode
```
kevinpollet, line 24: This description should be in a section for the next Helm chart version. 
Something like `v2.0 to v2.1`?

kevinpollet:
Is it a common practice to have a `global` key?
What are the benefits?

dtomcej:
> > We may want to introduce a global key similar to Traefik's config, to handle these values, as they seem to be piling up.
> 
> Is it a common practice to have a `global` key?
> What are the benefits?

I'm not sure if it is common practice, but it would group all the keys together for sanity and clarity

dtomcej:
Also, that would be a future issue if needed.

PR #623, opened 2020-06-05
Ignore ExternalName services

### LINKED ISSUE #618: Error logs with msg "empty args for matcher Host"
### What did you do?

setup maesh 1.2.0 in EKS.

### What did you see instead?

too many error logs with msg "empty args for matcher Host"

### Output of mesh log: (_1.2.0_)

```
maesh-mesh-4spq4 maesh-mesh 2020-05-31T02:33:08.392792516Z time="2020-05-31T02:33:08Z" level=error msg="empty args for matcher Host, []" entryPointName=http-5000 routerName=logging-aes-kibana-443@http
```

- EKS 1.19
- aes-kibana 443 is an external service of type = "ExternalName" in logging namespace.

jspdown:
## What does this PR do?

This PR update the IgnoreWrapper so it ignores Services of type `ExternalName`.

Fixes #618 

In the same time, I fixed an issue in the IgnoredWrapper on the way we handled ignored services. We weren't checking the type of the object, so if a pod had the same name as an ignored service we would have ignored it.

Some deadcode have also been removed.

### How to test it

- `make test`

### DIFF
--- a/pkg/k8s/ignore.go
+++ b/pkg/k8s/ignore.go
@@ -1,9 +1,8 @@
 package k8s
 
 import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 )
 
 // IgnoreWrapper holds namespaces and services to ignore.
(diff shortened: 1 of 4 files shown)


pkg/k8s/ignore.go
```diff
@@ -37,41 +36,36 @@ func (i *IgnoreWrapper) AddIgnoredApps(app ...string) {
 	i.Apps = append(i.Apps, app...)
 }
 
-// LabelSelector returns the labels.Selector image of the ignored object.
-func (i *IgnoreWrapper) LabelSelector() (labels.Selector, error) {
-	sel := labels.Everything()
-
-	r, err := labels.NewRequirement("app", selection.NotIn, i.Apps)
+// IsIgnored returns if the object events should be ignored.
```
Ullaakut, line 39:
```suggestion
// IsIgnored returns true if the object events should be ignored.
```

pkg/k8s/ignore.go
```diff
@@ -37,41 +36,36 @@ func (i *IgnoreWrapper) AddIgnoredApps(app ...string) {
...
+	// Is the app ignored?
+	if contains(i.Apps, pMeta.GetLabels()["app"]) {
 		return true
 	}
 
-	// Is the app ignored ?
-	if contains(i.Apps, obj.GetLabels()["app"]) {
-		return true
+	if svc, ok := obj.(*corev1.Service); ok {
+		// Is the object explicitly ignored?
```
Ullaakut, line 59: Aren't we checking specifically for services here?

```suggestion
		// Is the service explicitly ignored?
```

pkg/k8s/ignore.go
```diff
@@ -37,41 +36,36 @@ func (i *IgnoreWrapper) AddIgnoredApps(app ...string) {
...
-func (i *IgnoreWrapper) LabelSelector() (labels.Selector, error) {
-	sel := labels.Everything()
-
-	r, err := labels.NewRequirement("app", selection.NotIn, i.Apps)
+// IsIgnored returns true if the object events should be ignored.
+func (i *IgnoreWrapper) IsIgnored(obj interface{}) bool {
+	accessor, err := meta.Accessor(obj)
 	if err != nil {
-		return nil, err
+		return false
```
kevinpollet, line 43: In case of an error, we have to return `true` here, right?

```suggestion
		return true
```

PR #636, opened 2020-06-17
Stop accepting multiple TrafficSplits or both TrafficSplit and TrafficTarget on the same Service

### LINKED ISSUE #635: Topology can contains multiple TrafficSplits on the same Service
## Bug Report

Due to certain limitations in the SMI spec version implemented in Maesh, some configurations might leads to a topologies which contains connections that can't be followed.

The SMI version we are using doesn't support the `matches` attribute on TrafficSplits. In later version it's used for conditionally following a TrafficSplit. If you have multiple TrafficSplits, if a request doesn't pass the conditions of the first TrafficSplit we will check forf the conditions of the second one. If none passes the request will go to the pods which have a selector matching the service. 

As we don't support these conditions, there can be only one TrafficSplit attached to a service. However, Maesh topology will still contains the different TrafficSplits and Pods. Leading to an invalid/impossible Topology

### What did you do?

Deploy a service and 2 TrafficSplits and attach them to the service. 

A single TrafficSplit in the topology

### What did you see instead?

Got 2 TrafficSplit

version: master

jspdown:
## What does this PR do?

This PR:
- Checks if there is already a TrafficSplit on a service before adding a second one.
- Checks if there is already a TrafficSplit on a service before adding a TrafficTarget.
- Update the tests to cover these cases

Fixes #635 

### How to test it

* `make test`
* Manually by applying multiple traffic-split on the same service or a traffic-split + a traffic-target on the same service.

### DIFF
--- a/pkg/topology/builder.go
+++ b/pkg/topology/builder.go
@@ -59,7 +59,7 @@ func (b *Builder) Build(resourceFilter *mk8s.ResourceFilter) (*Topology, error) 
 	return topology, nil
 }
 
-// evaluateService evaluates the given service. It adds the Service to the topology and it's selected Pods.
+// evaluateService evaluates the given service. It adds the Service to the topology and its selected Pods.
 func (b *Builder) evaluateService(res *resources, topology *Topology, svc *corev1.Service) {
 	svcKey := Key{svc.Name, svc.Namespace}
 
@@ -104,10 +104,10 @@ func (b *Builder) evaluateTrafficTarget(res *resources, topology *Topology, tt *
 		}
 
 		svc, ok := topology.Services[svcKey]
-		// In the current version of the spec, the traffic will always go to a TrafficSplit. So we don't need to evaluate
-		// Pods if there's already a TrafficSplit on the Service.
+		// Since in the current version of the spec, the traffic will always go to a TrafficSplit, we don't need to evaluate
+		// Pods if there is already a TrafficSplit on the Service.
 		if ok && len(svc.TrafficSplits) > 0 {
-			b.Logger.Warnf("Service %q already has a TrafficSplit attached, TrafficTarget %q won't be evaluated on this service", svcKey, svcTTKey.TrafficTarget)
+			b.Logger.Warnf("Service %q already has a TrafficSplit attached, TrafficTarget %q will not be evaluated on this service", svcKey, svcTTKey.TrafficTarget)
 
 			return
 		}
@@ -177,7 +177,7 @@ func (b *Builder) buildTrafficTargetDestination(topology *Topology, tt *access.T
 func addSourceAndDestinationToPods(topology *Topology, sources []ServiceTrafficTargetSource, svcTTKey ServiceTrafficTargetKey) {
 	for _, source := range sources {
 		for _, podKey := range source.Pods {
-			// Skip pods which haven't been added to the topology.
+			// Skip pods which have not been added to the topology.
 			if _, ok := topology.Pods[podKey]; !ok {
 				continue
 			}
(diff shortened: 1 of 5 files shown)


pkg/topology/builder.go
```diff
@@ -44,16 +44,16 @@ func (b *Builder) Build(resourceFilter *mk8s.ResourceFilter) (*Topology, error)
...
-	for _, tt := range res.TrafficTargets {
-		b.evaluateTrafficTarget(res, topology, tt)
-	}
-
 	// Populate services with traffic-split definitions.
 	for _, ts := range res.TrafficSplits {
 		b.evaluateTrafficSplit(topology, ts)
 	}
 
+	// Populate services with traffic-target definitions.
```
Ullaakut, line 52: Why was it moved? Is there a specific reason?

pkg/topology/builder.go
```diff
@@ -106,13 +106,22 @@ func (b *Builder) evaluateTrafficTarget(res *resources, topology *Topology, tt *
 			Service:       svcKey,
 			TrafficTarget: Key{tt.Name, tt.Namespace},
 		}
-		topology.ServiceTrafficTargets[svcTTKey] = stt
 
 		svc, ok := topology.Services[svcKey]
+		// In the current version of the spec, the traffic will always go to a TrafficSplit. So we don't need to evaluate
```
Ullaakut, line 111:
```suggestion
		// Since in the current version of the spec, the traffic will always go to a TrafficSplit, we don't need to evaluate.
```

pkg/topology/builder.go
```diff
@@ -106,13 +106,22 @@ func (b *Builder) evaluateTrafficTarget(res *resources, topology *Topology, tt *
...
+		// Pods if there's already a TrafficSplit on the Service.
```
Ullaakut, line 112:
```suggestion
		// Pods if there is already a TrafficSplit on the Service.
```

Also I'm not sure I understand this sentence. It's missing a verb?

pkg/topology/builder.go
```diff
@@ -106,13 +106,22 @@ func (b *Builder) evaluateTrafficTarget(res *resources, topology *Topology, tt *
...
+		if ok && len(svc.TrafficSplits) > 0 {
+			b.Logger.Warnf("Service %q already has a TrafficSplit attached, TrafficTarget %q won't be evaluated on this service", svcKey, svcTTKey.TrafficTarget)
```
Ullaakut, line 114: (IMO we should write `will not` instead of `won't` in our logging. We're not having an informal conversation, the logs should stay format. but out of scope of this PR)

pkg/topology/builder_test.go
```diff
@@ -277,24 +279,61 @@ func TestTopologyBuilder_EvaluatesIncomingTrafficSplit(t *testing.T) {
 	svcKey := nn(svcB.Name, svcB.Namespace)
 	tsKeys := got.Services[svcKey].TrafficSplits
 
-	assert.Equal(t, 1, len(got.TrafficSplits[tsKeys[0]].Incoming))
-	assert.Equal(t, 1, len(got.TrafficSplits[tsKeys[1]].Incoming))
+	// Make sure only one TrafficSplit will be present.
```
Ullaakut, line 282:
```suggestion
	// Make sure the resulting Topology only has a single TrafficSplit.
```

pkg/topology/topology.go
```diff
@@ -108,7 +108,8 @@ func (k *ServiceTrafficTargetKey) UnmarshalJSON(data []byte) error {
 	return k.UnmarshalText(data)
 }
 
-// Topology holds the graph. Each Pods and services are nodes of the graph.
+// Topology holds the graph and represents the different path a request can follow. Each Pods and services are nodes
```
Ullaakut, line 111:
```suggestion
// Topology holds the graph and represents the different paths a request can follow. Each Pods and services are nodes
```

pkg/topology/builder.go
jspdown: It had to be moved since `evaluateTrafficTarget` needs to know if there's already a TrafficSplit on a service. So TrafficSplits have to be evaluated first.

pkg/topology/builder.go
jspdown: True, I often forget that :+1:

pkg/topology/builder.go
jspdown: This is not the begining of the sentence. This is why there was no `.` at the end of the previous line ;)

PR #637, opened 2020-06-17
Add customizable Grafana and Prometheus service object in Helm chart

### LINKED ISSUE #631: Do not expose Prometheus and Grafana services with a NodePort by default
### Proposal

Currently, `Prometheus` and `Grafana` services are exposed by default with a NodePort. Even if this behaviour seems desirable for development or to ease user onboarding, it is a best practice to not expose them by default.

One solution might be to add an `expose` option to the Helm chart to configure whenever or not those services should be exposed.

### Background

Forum post: https://community.containo.us/t/prometheus-and-grafana-of-type-nodeport

walbertus:
## What does this PR do?

This PR:

- Add customization for Grafana and Prometheus service object in helm chart including:
  - Service type
  - Service port
  - Service NodePort

Fixes #631 

### How to test it

* Go to `./helm/chart/maesh`
* Update desired `values.yaml`
* `helm install maesh .`

### DIFF
--- a/helm/chart/maesh/Chart.yaml
+++ b/helm/chart/maesh/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v2
 name: maesh
-version: 2.0.1
+version: 2.0.0
 appVersion: v1.2.0
 description: Maesh - Simpler Service Mesh
 keywords:
@@ -25,5 +25,5 @@ dependencies:
     version: 0.0.3
     condition: tracing.deploy
   - name: metrics
-    version: 0.1.0
+    version: 0.0.6
     condition: metrics.deploy
--- a/helm/chart/maesh/Chart.yaml
+++ b/helm/chart/maesh/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v2
 name: maesh
-version: 2.0.0
+version: 2.0.1
 appVersion: v1.2.0
 description: Maesh - Simpler Service Mesh
 keywords:
@@ -25,5 +25,5 @@ dependencies:
     version: 0.0.3
     condition: tracing.deploy
   - name: metrics
-    version: 0.0.6
+    version: 0.1.0
     condition: metrics.deploy
--- a/helm/chart/maesh/charts/metrics/Chart.yaml
+++ b/helm/chart/maesh/charts/metrics/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 description: A metrics Helm chart for Kubernetes
 name: metrics
-version: 0.1.0
+version: 0.0.6
 appVersion: 0.0.1
 tillerVersion: ">=2.7.2"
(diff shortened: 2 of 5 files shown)


helm/chart/maesh/charts/metrics/values.yaml
```diff
@@ -4,6 +4,10 @@ image:
   configmapReload: jimmidyson/configmap-reload:v0.2.2
 
 grafana:
+  service:
+    type: ClusterIP
+    port: 3000
+  # nodePort: null
```
Ullaakut, line 10: Please document the `nodePort` field, like other commented fields are documented.

walbertus: Thanks for the review, updated to include comments for nodePort

helm/chart/maesh/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 apiVersion: v2
 name: maesh
-version: 2.0.0
+version: 2.0.1
```
kevinpollet, line 4: The Helm chart version must be updated in the Release PR (the Helm chart is not deployed or released on each commit like it is done for Traefik).

```suggestion
version: 2.0.0
```

helm/chart/maesh/Chart.yaml
```diff
@@ -25,5 +25,5 @@ dependencies:
     version: 0.0.3
     condition: tracing.deploy
   - name: metrics
-    version: 0.0.6
+    version: 0.1.0
```
kevinpollet, line 28: See my previous comment.

```suggestion
    version: 0.0.6
```

helm/chart/maesh/charts/metrics/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 description: A metrics Helm chart for Kubernetes
 name: metrics
-version: 0.0.6
+version: 0.1.0
```
kevinpollet, line 4: See my previous comment.

```suggestion
version: 0.0.6
```

helm/chart/maesh/charts/metrics/values.yaml
```diff
@@ -4,6 +4,15 @@ image:
...
 
 grafana:
+  service:
+    type: ClusterIP
+    port: 3000
+
+    ## Specify the nodePort value for the NodePort service types.
+    ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#nodeport
+    ##
+    # nodePort: null
```
kevinpollet, line 14: I think that it's better to use a dummy `nodePort` for the example instead of `null`.

```suggestion
    # nodePort: 30030
```

helm/chart/maesh/charts/metrics/values.yaml
```diff
@@ -18,6 +27,15 @@ grafana:
...
 
 prometheus:
+  service:
+    type: ClusterIP
+    port: 9090
+
+    ## Specify the nodePort value for the NodePort service types.
+    ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#nodeport
+    ##
+    # nodePort: null
```
kevinpollet, line 37: I think that it's better to use a dummy `nodePort` for the example instead of null.

```suggestion
    # nodePort: 30090
```

PR #656, opened 2020-07-01
Create optional kube-dns configmap if needed

### LINKED ISSUE #652: KubeDNS patch not working on EKS cluster created with kops
### What did you do?

Deployed Maesh on EKS cluster created with `kops`.

A working Maesh installation.

### What did you see instead?

The init container of the Maesh controller is in Error.

### Output of controller log:

```
time="2020-07-01T07:16:49Z" level=info msg="Configuration loaded from flags."
2020/07/01 07:16:49 command prepare error: unable to configure DNS: unable to configure KubeDNS: configmaps "kube-dns" not found
```

kevinpollet:
## What does this PR do?

This PR:

- Cleanup the KubeDNS patching code.
- Fixes the KubeDNS patching code to create the `kube-dns` ConfigMap if it is optional and not yet created. For example, in a Kubernetes cluster deployed with `kops`, the `kube-dns` ConfigMap is optional and not created by default.

Fixes #652

### How to test it

* make test
* make test-integration

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -55,7 +55,7 @@ func NewClient(logger logrus.FieldLogger, kubeClient kubernetes.Interface) *Clie
 	}
 }
 
-// CheckDNSProvider checks that the DNS provider that is deployed in the cluster is supported and returns it.
+// CheckDNSProvider checks that the DNS provider deployed in the cluster is supported and returns it.
 func (c *Client) CheckDNSProvider() (Provider, error) {
 	c.logger.Info("Checking DNS provider")
 
@@ -169,7 +169,7 @@ func (c *Client) ConfigureCoreDNS(coreDNSNamespace, clusterDomain, maeshNamespac
 }
 
 func (c *Client) patchCoreDNSConfig(deployment *appsv1.Deployment, clusterDomain, maeshNamespace string) (*corev1.ConfigMap, error) {
-	customConfigMap, err := c.getConfigMap(deployment.Namespace, "coredns-custom")
+	customConfigMap, err := c.getConfigMap(deployment, "coredns-custom")
 
 	// For AKS the CoreDNS config have to be added to the coredns-custom ConfigMap.
 	// See https://docs.microsoft.com/en-us/azure/aks/coredns-custom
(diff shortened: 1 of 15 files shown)


pkg/dns/dns.go
```diff
@@ -427,49 +431,43 @@ func (c *Client) unpatchCoreDNSConfig(deployment *appsv1.Deployment) (*corev1.Co
...
 func (c *Client) RestoreKubeDNS() error {
-	deployment, err := c.kubeClient.AppsV1().Deployments(metav1.NamespaceSystem).Get("kube-dns", metav1.GetOptions{})
+	kubeDNSDeployment, err := c.kubeClient.AppsV1().Deployments(metav1.NamespaceSystem).Get("kube-dns", metav1.GetOptions{})
 	if err != nil {
 		return err
 	}
 
 	// Get the currently loaded KubeDNS ConfigMap.
-	configMap, err := c.getKubeDNSConfigMap(deployment)
+	configMap, err := c.getConfigMap(kubeDNSDeployment.Namespace, "kube-dns")
```
jspdown, line 440: In most situation, assuming there will be a `kube-dns` ConfigMap in the KubeDNS deployment namespace is true. The restore logic for getting the ConfigMap is not the same as the patch logic. Not a blocker but I think it would be more robust to also using the iteration on volume logic here as well.

pkg/dns/dns.go
```diff
@@ -245,106 +244,111 @@ maesh:53 {
...
+func (c *Client) getOrCreateKubeDNSConfigMap(deployment *appsv1.Deployment) (*corev1.ConfigMap, error) {
+	for _, volume := range deployment.Spec.Template.Spec.Volumes {
+		if volume.ConfigMap == nil {
+			continue
+		}
 
-	if kubeConfigMap.Data == nil {
-		kubeConfigMap.Data = make(map[string]string)
-	}
+		if volume.ConfigMap.Name != "kube-dns" {
```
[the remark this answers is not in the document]
kevinpollet: I thought exactly the same, and I don't have a strong opinion. My thinking is that in Kubernetes documentations they always talk about a `kubedns` configmap, so that's why I've done it.

You're right, I think this should be configurable.

pkg/dns/dns.go
```diff
@@ -134,7 +133,7 @@ func (c *Client) kubeDNSMatch() (bool, error) {
 	}
 
 	if err != nil {
-		return false, fmt.Errorf("unable to get KubeDNS deployment in namesapce %q: %w", metav1.NamespaceSystem, err)
+		return false, fmt.Errorf("unable to get KubeDNS deployment in namespace %q: %w", metav1.NamespaceSystem, err)
 	}
 
 	c.logger.Info("KubeDNS match")
```
jspdown, line 139: We should consider changing this to a `Debug` log level. Same for:
- `Checking CoreDNS`
- `Checking KubeDNS`
- `CoreDNS match`

Ideally, the INFO output of the prepare command should be:

```
> DNS provider detected: KubeDNS
> Patching KubeDNS
> Restarting "coredns" pods
> Restarting "kube-dns" pods
```

You don't have to do it in this PR. But we need to fix this at some point, the output of the prepare command needs to be improved.

kevinpollet: You're right, can you create an issue for that?

PR #665, opened 2020-07-08
Prepare Release v1.3.1

kevinpollet:
🚢


Ullaakut:
In the future when fixing a release, please document what was wrong with the previous one 🙏 Since the v1.3.1's changelog is just `Prepare release`, it's really not clear at the moment

PR #679, opened 2020-07-20
Use Kubernetes semver to add topology keys if supported

### LINKED ISSUE #678: Kubernetes Minor version parsing could fail
## Bug Report

As said in the Kubernetes code the `Minor` version returned by `.Discovery().ServerVersion().Major` could potentially contain a `+`. Therefore, the parsing of the `Minor` version will fail and the topology annotations will not be added, even if the Kubernetes cluster is supporting it.

Maesh should compare the Kubernetes version by using the `semver` string or by removing the `+`.

kevinpollet:
This PR:

- Adds [go-version](https://github.com/hashicorp/go-version#readme) dependency.
- Uses the Kubernetes server `semver` to check if topology keys are supported.
- Fixes a parsing error when the Kubernetes server `minor` version contains a `+`, see https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/version/base.go#L43

Fixes #678.

### How to test it

* make test

### DIFF
--- a/pkg/controller/service.go
+++ b/pkg/controller/service.go
@@ -3,9 +3,9 @@ package controller
 import (
 	"errors"
 	"fmt"
-	"strconv"
 
 	"github.com/containous/maesh/pkg/annotations"
+	"github.com/hashicorp/go-version"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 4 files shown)


pkg/controller/service.go
```diff
@@ -87,7 +89,12 @@ func (s *ShadowServiceManager) CreateOrUpdate(svc *corev1.Service) (*corev1.Serv
 	}
 
 	// If the kubernetes server version is 1.17+, then use the topology key.
-	if major, minor := parseKubernetesServerVersion(s.kubeClient); major == 1 && minor >= 17 {
+	serverVersion, err := getServerVersion(s.kubeClient)
```
jspdown, line 92: What do you think about doing it only once in the `NewShadowServiceManager`? `kubeClient.Discovery().ServerVersion()` is making a request on the server each time we call it.

kevinpollet: Why not 🙂

kevinpollet: After trying to do it, it appears that we have to handle the error through all the construction chain, which is not ideal. My thinking is that it's better to keep it like that for now.

PR #703, opened 2020-08-17
Update SMI resource versions

jspdown:
## What does this PR do?

This PR:
- Update specs.smi-spec.io version from `v1alpha1` to `v1alpha3`
- Update access.smi-spec.io version from `v1alpha1` to `v1alpha2`
- Update split.smi-spec.io version from `v1alpha2` to `v1alpha3`

This PR updates the version but doesn't provide support for Header based filtering on TrafficTarget and filtering on TrafficSplit. These features will be implemented in separate PRs.

### How to test it

* `make test`
* `make test-integration`


helm/chart/maesh/crds/smi-trafficsplits.yaml
```diff
@@ -5,21 +5,67 @@ metadata:
   name: trafficsplits.split.smi-spec.io
 spec:
   group: split.smi-spec.io
-  version: v1alpha2
```
kevinpollet, line 8: To be consistent with the previous CRD file names, we should rename this file to `smi-split.yaml`.
wdyt?

helm/chart/maesh/crds/smi-spec.yaml
```diff
@@ -5,27 +5,92 @@ metadata:
   name: httproutegroups.specs.smi-spec.io
 spec:
   group: specs.smi-spec.io
-  version: v1alpha1
```
kevinpollet, line 8: To be consistent with the CRD name I think we should rename this file to `smi-specs.yaml`.
wdyt?

integration/testdata/smi/crds/smi-spec.yaml
```diff
@@ -5,27 +5,92 @@ metadata:
   name: httproutegroups.specs.smi-spec.io
 spec:
   group: specs.smi-spec.io
-  version: v1alpha1
```
kevinpollet, line 8: See my above comment.

integration/testdata/smi/crds/smi-trafficsplits.yaml
```diff
@@ -5,21 +5,67 @@ metadata:
   name: trafficsplits.split.smi-spec.io
 spec:
   group: split.smi-spec.io
-  version: v1alpha2
```
kevinpollet, line 8: See my above comment.

PR #704, opened 2020-08-18
Speed up integration tests and improve CoreDNSSuite

jspdown:
## What does this PR do?

This PR speeds up running integration tests in parallel and improve the CoreDNSSuite:

- Improve performances of the CoreDNSSuite:

This test suite has been refactored so we no longer just test that we didn't received and error when patching the corefile but instead also check that the patched corefile can be used to resolve a `.maesh` url.

Some code has also been changed in the dns package to improve the stability of the coredns patching.
Starting from now, even if the corefile/custom corefile has already been patched we will re-patch it.
Since version 1.7.0 of CoreDNS, the `upstream` keyword has been removed, so, if a config was already patched for a previous version or CoreDNS but then the user upgraded it, the corefile would have not been repatched. This commit fix it. This fix also helped us speeding up the CoreDNS tests has we no longer need to re-install a complete CoreDNS before each test.

The coredns/coredns:1.2.6 image has been removed as we are not testing anymore that we don't support it.

- Speed up integration tests when run in parallel:

Instead of pulling all the images required by all test before doing anything we are now delegating the pull operation to the `SetUpSuite`. Starting from now, each test is now responsible for pulling its images.

Tests that don't require many image will benefits a lot from this change.

### How to test it

* Run integration tests sequencially using `make test-integration`
* Let the CI run integration tests in parallel

## Additional Notes

Another PR will follow to give us more confidence in the ACLDisableSuite and ACLEnabledSuite by actually making request in the mesh instead of comparing dynamic configuration.


integration/integration_test.go
```diff
@@ -61,42 +60,13 @@ func Test(t *testing.T) {
...
-		}
-	}
-
 	check.TestingT(t)
 }
 
 type image struct {
-	name string
-	pull bool
+	repository string
```
kevinpollet, line 67: Here I would have kept the property `name` as it is usally the way to reference a docker image.
wdyt?

pkg/dns/dns.go
```diff
@@ -511,3 +464,66 @@ func getConfigMapVolumeSource(deployment *appsv1.Deployment, name string) (*core
...
+		strings.Replace(clusterDomain, ".", "\\.", -1),
+		traefikMeshNamespace,
+		blockHeader,
+		blockTrailer,
+		upstream,
+		domain,
+	)
+
+	return config + stubDomain
+}
```
kevinpollet, line 511:
```suggestion
}

```

PR #705, opened 2020-08-18
Add support for header-based filtering on TrafficTargets

jspdown:
## What does this PR do?

This PR adds support for header-based filtering on TrafficTargets. As described in the specification, an HTTPRouteGroup can define a set of headers that must match to pass through.

https://github.com/servicemeshinterface/smi-spec/blob/master/apis/traffic-specs/v1alpha3/traffic-specs.md#http-header-filters

```
kind: HTTPRouteGroup
metadata:
  name: the-routes
  namespace: default
spec:
  matches:
  - name: android-insiders
    headers:
    - user-agent: ".*Android.*"
    - cookie: "^(.*?;)?(type=insider)(;.*)?$"
```

### How to test it

* `make test`
* `make test-integration`
* Install maesh and play with this feature. Instruction on how to use this new filter can be found in the specification

### DIFF
--- a/pkg/provider/rule.go
+++ b/pkg/provider/rule.go
@@ -76,7 +76,7 @@ func appendMethodFilter(matchParts []string, match *specs.HTTPMatch) []string {
 }
 
 func appendHeaderFilter(matchParts []string, match *specs.HTTPMatch) []string {
-	var rules []string
+	rules := make([]string, 0, len(match.Headers))
 
 	for name, value := range match.Headers {
 		rules = append(rules, fmt.Sprintf("HeadersRegexp(`%s`, `%s`)", name, value))
--- a/pkg/provider/rule.go
+++ b/pkg/provider/rule.go
@@ -21,6 +21,9 @@ func buildTrafficTargetRule(tt *topology.ServiceTrafficTarget) string {
 			// Handle Method filtering.
 			matchParts = appendMethodFilter(matchParts, match)
 
+			// Handle Header filtering.
+			matchParts = appendHeaderFilter(matchParts, match)
+
 			// Conditions within a HTTPMatch must all be fulfilled to be considered valid.
 			if len(matchParts) > 0 {
 				matchCond := strings.Join(matchParts, " && ")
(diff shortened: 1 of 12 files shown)


pkg/provider/rule.go
```diff
@@ -72,6 +75,20 @@ func appendMethodFilter(matchParts []string, match *specs.HTTPMatch) []string {
...
+
+	for name, value := range match.Headers {
+		rules = append(rules, fmt.Sprintf("HeadersRegexp(`%s`, `%s`)", name, value))
+	}
+
+	if len(rules) > 0 {
+		matchParts = append(matchParts, strings.Join(rules, " && "))
+	}
+
+	return matchParts
```
kevinpollet, line 89:
```suggestion
	return append(matchParts, strings.Join(rules, " && "))
```

pkg/provider/rule.go
```diff
@@ -72,6 +75,20 @@ func appendMethodFilter(matchParts []string, match *specs.HTTPMatch) []string {
...
 }
 
+func appendHeaderFilter(matchParts []string, match *specs.HTTPMatch) []string {
+	var rules []string
+
+	for name, value := range match.Headers {
+		rules = append(rules, fmt.Sprintf("HeadersRegexp(`%s`, `%s`)", name, value))
+	}
+
+	if len(rules) > 0 {
```
kevinpollet, line 85: This `if` block is not necessary.

pkg/provider/rule.go
```diff
@@ -72,6 +75,20 @@ func appendMethodFilter(matchParts []string, match *specs.HTTPMatch) []string {
 	return matchParts
 }
 
+func appendHeaderFilter(matchParts []string, match *specs.HTTPMatch) []string {
+	var rules []string
+
+	for name, value := range match.Headers {
```
kevinpollet, line 81: Questions:

- Is the `httpHeaders` type necessary?
- Should we rename `ServiceTrafficTarget#Specs` to `ServiceTrafficTarget#Rules`?
- Do you think that it could make sense to create the `rules` array with the right size? 

```
rules := make([]string, len(match.Headers))
```

kevinpollet (changes requested):

PR #709, opened 2020-08-21
Add support for TrafficSplit filtering

### LINKED ISSUE #598: Support Traffic Split version v1alpha3
### Proposal

Support the version `v1alpha3` of the `Traffic Split` API from the [SMI specification](https://github.com/servicemeshinterface/smi-spec/blob/master/apis/traffic-split/v1alpha3/traffic-split.md) (the current implemented version is `v1alpha2`).

As explained in the specification, this version introduces a `matches` property allowing to define HTTP header filters to route a specific user segment to a backend while all the other users, not belonging to that segment, will be routed to the default service backend.

jspdown:
## What does this PR do?

This PR:
- Add new property `Rules` in the `TrafficSplit` struct
- Build SMI specs rules for each `TrafficSplit` in the TopologyBuilder.
- Update topology tests

Fixes #598 

### How to test it
- `make test`
- Look at the specification (split.v1alpha3) and play with the new feature.


kevinpollet (approved):
LGTM 🚢

PR #710, opened 2020-08-21
Always update CoreDNS deployment in prepare phase

jspdown:
## What does this PR do?

This PR:
- Checks if the corefile has changed before updating the coredns deployment

### How to test it

* make test


kevinpollet (approved):
LGTM 🚢

PR #713, opened 2020-08-24
Fix controller signal handling

### LINKED ISSUE #702: Controller loop does not exit and cannot be stopped if SMI CRDs are missing
### What did you do?

Started the controller binary from the command-line with the `MAESH_KUBECONFIG`env variable which targets a Kubernetes cluster without the SMI CRDs.

A running controller which:

- exists after an amount of time
- can be stopped by sending a `SIGINT` or `SIGTERM` signal or which exits after 

### What did you see instead?

The control loop cannot be stopped.

jspdown:
## What does this PR do?

Fix a bug in `cmd/maesh` package where a WaitGroup was getting stuck forever. If the `controller` failed after the program received a signal (e.g. CTRL-C), the error couldn't be sent to the error chan and would block forever.

Fixes #702 

### How to test it

* Run locally the controller with the `MAESH_KUBECONFIG` variable set to a k8s cluster with no SMI CRDs. After few seconds, send a signal. It should stop immediately.

### DIFF
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -7,6 +7,23 @@ import (
 	"syscall"
 )
 
+// ContextWithStopChan creates a context canceled when the given stopCh receives a message
+// or get closed.
+func ContextWithStopChan(ctx context.Context, stopCh <-chan struct{}) context.Context {
+	ctx, cancel := context.WithCancel(ctx)
+
+	go func() {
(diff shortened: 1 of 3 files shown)


kevinpollet (approved):
LGTM 🚢

PR #714, opened 2020-08-25
Check SMI CRDs version before starting the controller

### LINKED ISSUE #711: Prepare command shouldn't start informers
## Feature Request

The prepare command does two things:
- Start informers and wait for them to sync up
- Patch DNS

### Proposal

The first step is not necessary and should be removed. The original idea was to check in the prepare command if the right CRDs are in place and that we can fetch resources before starting the controller. If it doesn't, the prepare command would fail. This make no difference compared to the controller failing. 

I suggest we remove this check, this would allow a faster start time of the controller, and would allow us to remove the `--acl` flag from this command. 
However, it could still be interesting to check if the right CRDs for SMI are installed, with the right version. But we shouldn't have to start informers for this.

### Workarounds

It can be kept like that, it's just not optimal.

jspdown:
## What does this PR do?

In order to improve the user experience when incompatible SMI CRDs versions are installed (outdated or not yet supported), this PR introduce some changes on the prepare command. 

In the past we use to start informers and start collecting the different SMI objects already present on the cluster. If SMI CRDs were not installed or an unsupported version was we would get the following errors many times:

```
E0824 17:40:01.430644   26974 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.6/tools/cache/reflector.go:125: Failed to list *v1alpha3.TrafficSplit: the server could not find the requested resource (get trafficsplits.split.smi-spec.io)
E0824 17:40:01.430644   26974 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.6/tools/cache/reflector.go:125: Failed to list *v1alpha3.TrafficSplit: the server could not find the requested resource (get trafficsplits.split.smi-spec.io)
E0824 17:40:01.430644   26974 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.6/tools/cache/reflector.go:125: Failed to list *v1alpha3.TrafficSplit: the server could not find the requested resource (get trafficsplits.split.smi-spec.io)
E0824 17:40:01.430644   26974 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.6/tools/cache/reflector.go:125: Failed to list *v1alpha3.TrafficSplit: the server could not find the requested resource (get trafficsplits.split.smi-spec.io)
E0824 17:40:01.430644   26974 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.6/tools/cache/reflector.go:125: Failed to list *v1alpha3.TrafficSplit: the server could not find the requested resource (get trafficsplits.split.smi-spec.io)
```

This PR introduce a simpler solution based on the `Discovery` API of the kubernetes sdk to discover which resource groups are installed and what are there versions. 

Fixes #711

### How to test it

* Install Maesh with Helm using the `--skip-crds` flag: The controller should fail and you should see a message in the init container logs telling you that the SMI CRDs are missing.
* Install Maesh normally: This should work fine
* Create a cluster, install an old version of the SMI CRDs (e.g. specs v1alpha) and install Maesh using Helm with the `--skip-crds` you should see an error in the logs too.


cmd/prepare/prepare.go
```diff
@@ -50,12 +51,26 @@ func prepareCommand(pConfig *cmd.PrepareConfiguration) error {
...
+
+	dnsProvider, err = dnsClient.CheckDNSProvider(ctx)
+	if err != nil {
+		return fmt.Errorf("unable to find suitable DNS provider: %w", err)
+	}
+
+	switch dnsProvider {
+	case dns.CoreDNS:
+		if err := dnsClient.ConfigureCoreDNS(ctx, metav1.NamespaceSystem, pConfig.ClusterDomain, pConfig.Namespace); err != nil {
+			return fmt.Errorf("unable to configure CoreDNS: %v", err)
```
kevinpollet, line 68:
```suggestion
			return fmt.Errorf("unable to configure CoreDNS: %w", err)
```

cmd/prepare/prepare.go
```diff
@@ -50,12 +51,26 @@ func prepareCommand(pConfig *cmd.PrepareConfiguration) error {
...
+		}
+	case dns.KubeDNS:
+		if err := dnsClient.ConfigureKubeDNS(ctx, pConfig.ClusterDomain, pConfig.Namespace); err != nil {
+			return fmt.Errorf("unable to configure KubeDNS: %v", err)
```
kevinpollet, line 72:
```suggestion
			return fmt.Errorf("unable to configure KubeDNS: %w", err)
```

pkg/k8s/client.go
```diff
@@ -150,3 +150,18 @@ func buildSmiSplitClient(log logrus.FieldLogger, config *rest.Config) (*splitcli
 
 	return client, nil
 }
+
+// NewKubernetesClient creates a new Kubernetes client.
+func NewKubernetesClient(log logrus.FieldLogger, masterURL, kubeConfig string) (*kubernetes.Clientset, error) {
```
kevinpollet, line 155: For now, I would just keep the `NewClient` method and we will revisit this part later.

pkg/k8s/smi.go
```diff
@@ -0,0 +1,57 @@
...
+	"strings"
+
+	access "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/access/v1alpha2"
+	specs "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/specs/v1alpha3"
+	split "github.com/servicemeshinterface/smi-sdk-go/pkg/apis/split/v1alpha3"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes"
+)
+
+// CheckSMIVersion checks if the SMI CRDs versions installed matches what Maesh supports.
```
kevinpollet, line 15:
```suggestion
// CheckSMIVersion checks if the SMI CRDs versions installed match the supported versions.
```

pkg/k8s/smi.go
```diff
@@ -0,0 +1,57 @@
...
+	var errs []string
+
+	for _, requiredGroup := range requiredGroups {
+		var found bool
+
+		var version string
+
+		for _, group := range serverGroups.Groups {
+			if requiredGroup.Group == group.Name {
+				found = true
```
kevinpollet, line 40: I'm not sure we need the found `bool` here and I think we can add a break.

kevinpollet (changes requested):

pkg/k8s/smi.go
jspdown: I could definitely live without this variable but the issue was to report every CRD with a mismatched version.

kevinpollet (approved):
LGTM 🚢

PR #715, opened 2020-08-26
Fix timeout on start informers

jspdown:
## What does this PR do?

This PR fixes an issue introduced in which causes the controller to stop after 10s. 

### How to test it

* Start Maesh and apply a manifest containing a service. You should see a shadow-service in maesh namespace.


kevinpollet (approved):
LGTM 👍

PR #716, opened 2020-08-27
Remove unnecessary RBAC rules for controller and proxy

kevinpollet:
## What does this PR do?

This PR:

- Removes unnecessary RBAC rules for controller and proxy
- Removes dead code

### How to test it

* make test-integration


pkg/k8s/client_mock.go
```diff
@@ -89,49 +88,29 @@ func NewClientMock(testingT *testing.T, stopCh <-chan struct{}, path string, acl
...
-	namespaceInformer := c.informerFactory.Core().V1().Namespaces().Informer()
-	trafficSplitInformer := c.splitInformerFactory.Split().V1alpha3().TrafficSplits().Informer()
-
-	podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
-	serviceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
-	endpointsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
-	namespaceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
-	trafficSplitInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{})
-
 	c.PodLister = c.informerFactory.Core().V1().Pods().Lister()
```
jspdown, line 97: We could also get rid of the listers. They doesn't seem to be used.

jspdown (approved):

PR #717, opened 2020-08-28
Improve prepare command logs

### LINKED ISSUE #660: Improve prepare command logs
### Proposal

Re-work the log messages of the `prepare` command. The INFO log level should be used only for information the user needs to know, such as choices made by the command and pod restarts. 

I suggest we update it so it looks something like this:

```
> [INFO] DNS provider detected: KubeDNS
> [INFO] Patching KubeDNS
> [INFO] Restarting "coredns" pods
> [INFO] Restarting "kube-dns" pods
```

### Background

The `prepare` command output could be improved to log in INFO mode only what we the user really needs to know. The current output is the following:

```
DEBU[0000] Building Kubernetes Client...                
DEBU[0000] Building SMI Access Client...                
DEBU[0000] Building SMI Specs Client...                 
DEBU[0000] Building SMI Split Client...                 
INFO[0000] Checking DNS provider                        
INFO[0000] Checking CoreDNS                             
DEBU[0000] CoreDNS deployment does not exist in namespace "kube-system" 
INFO[0000] Checking KubeDNS                             
INFO[0000] KubeDNS match                                
DEBU[0000] Patching KubeDNS                             
DEBU[0000] Getting CoreDNS service IP                   
DEBU[0000] Patching KubeDNS configmap with IP10.35.240.80 
DEBU[0000] Patching CoreDNS configmap                   
DEBU[0000] Patching CoreDNS                             
DEBU[0000] Patching CoreDNS configmap                   
INFO[0000] Restarting "coredns" pods                    
INFO[0000] Restarting "kube-dns" pods
```
### Workarounds

Keep it as it is.

jspdown:
## What does this PR do?

This PR reworks the log messages of the prepare command. So logs which are considered useful for debugging have been move to the DEBUG level. Some context has been added to the logs.

Fixes #660 

### How to test it

* run `maesh prepare` a local cluster, try with the `--logLevel` option

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -216,7 +216,7 @@ func (c *Client) getCoreDNSVersion(deployment *appsv1.Deployment) (*goversion.Ve
 		return goversion.NewVersion(parts[len(parts)-1])
 	}
 
-	return nil, fmt.Errorf("unable to get CoreDNS container in deployment %q in namespace %q", deployment.Namespace, deployment.Name)
+	return nil, fmt.Errorf("unable to get CoreDNS container in deployment %q in namespace %q", deployment.Name, deployment.Namespace)
 }
 
 // ConfigureKubeDNS patches the KubeDNS configuration for Maesh.
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -52,7 +52,7 @@ func NewClient(logger logrus.FieldLogger, kubeClient kubernetes.Interface) *Clie
 
 // CheckDNSProvider checks that the DNS provider deployed in the cluster is supported and returns it.
 func (c *Client) CheckDNSProvider(ctx context.Context) (Provider, error) {
-	c.logger.Info("Checking DNS provider")
+	c.logger.Debug("Detecting DNS provider...")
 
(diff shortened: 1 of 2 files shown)


pkg/dns/dns.go
```diff
@@ -212,30 +216,30 @@ func (c *Client) getCoreDNSVersion(deployment *appsv1.Deployment) (*goversion.Ve
 		return goversion.NewVersion(parts[len(parts)-1])
 	}
 
-	return nil, fmt.Errorf("unable to get CoreDNS container in deployment %q/%q", deployment.Namespace, deployment.Name)
+	return nil, fmt.Errorf("unable to get CoreDNS container in deployment %q in namespace %q", deployment.Namespace, deployment.Name)
```
kevinpollet, line 219:
```suggestion
	return nil, fmt.Errorf("unable to get CoreDNS container in deployment %q in namespace %q", deployment.Name, deployment.Namespace)
```

kevinpollet (changes requested):

pkg/dns/dns.go
jspdown: :eagle: Good catch!

kevinpollet (approved):
LGTM 🚢

PR #742, opened 2020-09-25
Remove support of maesh env and config files

kevinpollet:
## What does this PR do?

This PR removes the support of maesh env and config files.

### How to test it

* make test
* make test-integration

### DIFF
--- a/cmd/loaders.go
+++ b/cmd/loaders.go
@@ -16,54 +16,36 @@ import (
 
 const (
 	meshPrefix        = "MESH_"
-	maeshPrefix       = "MAESH_"
 	traefikMeshPrefix = "TRAEFIK_MESH_"
 )
 
 // EnvLoader loads a configuration from all the environment variables.
 type EnvLoader struct{}
 
-// Load loads the command's configuration from the environment variables prefixed with "TRAEFIK_MESH_" or "MAESH_".
-// The "MAESH_" prefix is deprecated and will be removed in the next major release.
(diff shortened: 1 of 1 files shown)


cmd/loaders.go
```diff
@@ -16,38 +16,30 @@ import (
...
+	traefikMeshVars := env.FindPrefixedEnvVars(os.Environ(), traefikMeshPrefix, cmd.Configuration)
 
-	traefikMeshVars := findAndNormalizePrefixedEnvVars(traefikMeshPrefix, cmd.Configuration)
-	maeshVars := findAndNormalizePrefixedEnvVars(maeshPrefix, cmd.Configuration)
+	var meshVars []string
 
-	if len(maeshVars) > 0 && len(traefikMeshVars) > 0 {
-		return false, fmt.Errorf("environment variable prefixed by %q cannot be mixed with variable prefixed by %q", maeshPrefix, traefikMeshPrefix)
+	for _, v := range traefikMeshVars {
+		meshVars = append(meshVars, strings.Replace(v, traefikMeshPrefix, meshPrefix, 1))
```
jspdown, line 34: Out of curiosity, why do we need to replace `traefikMeshPrefix` by `meshPrefix`? What would be the difference with just passing `traefikMeshPrefix` to `env.Decode`?

kevinpollet: Because it is currently not possible to have a prefix with multiple "_" in paerser.

dtomcej (approved):
LGTM
:shipit:

jspdown (approved):
LGTM :rocket:

PR #751, opened 2020-09-29
Use standard labels

jspdown:
## What does this PR do?

This PR changes the labels used in the resources managed by Traefik Mesh. Labels app and type have been replaced by:

- `app.kubernetes.io/name`
- `app.kubernetes.io/component`
- `app.kubernetes.io/part-of`

Kubernetes reference: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

This PR also remove some test files that are not being used.

### How to test it

* `make test`
* `make test-integration`


pkg/k8s/labels.go
```diff
@@ -0,0 +1,55 @@
...
+	AppName = "traefik-mesh"
+
+	// ComponentProxy is component of type proxy.
+	ComponentProxy = "proxy"
+	// ComponentShadowService is component of type shadow-service.
+	ComponentShadowService = "shadow-service"
+)
+
+// ShadowServiceLabels are the labels of a shadow service.
+var ShadowServiceLabels = map[string]string{
```
kevinpollet, line 26: WDYT of using a function here to avoid a mutable shared var and to be consistent with the selector functions. Something like:

```suggestion
func ShadowServiceLabels() map[string]string
```

kevinpollet (commented):
A minor suggestion :)

Good work 👍

SantoDE (approved):
LGTM

PR #757, opened 2020-10-01
Remove configuration file support

### LINKED ISSUE #755: Remove file loader and --configFile flag
### Proposal

Remove the `--configFile` flag. As there's not a lot of options we can give to Traefik Mesh and since all of these options are following a flat structure, there is no need to allow configuring it using a configuration file. 

### Workarounds

Just keep it.

kevinpollet:
## What does this PR do?

This PR:

- Removes the support of configuration file.
- Renames the configuration param in command functions.

Fixes #755

### How to test it

* make test
* make test-integration


jspdown (approved):
LGTM :+1:

PR #767, opened 2020-10-19
Deny HTTP traffic when no HTTPRouteGroup is given

### LINKED ISSUE #747: Traffic should be denied if TrafficTarget has no rules
### What did you do?

Deployed a `TrafficTarget` without `rules`.

Here is the deployed application: [traffic-access.zip](https://github.com/traefik/mesh/files/5292390/traffic-access.zip).

As explained in the [traffic-access specification](https://github.com/servicemeshinterface/smi-spec/blob/master/apis/traffic-access/v1alpha2/traffic-access.md) traffic should be denied.

> Any pod which attempts to connect and is not in the defined list of sources will be denied. Any pod which is in the defined list but attempts to connect on a route which is not in the list of `TrafficSpecs` will be denied.

### What did you see instead?

Traffic is allowed.

jspdown:
## What does this PR do?

This PR:
- Forbid HTTP traffic if the TrafficTarget doesn't contain an HTTPRouteGroup (according to the specification)
- Remove the HTTPMatches field of the ServiceTrafficTarget struct. Instead, the HTTPRouteGroup.Spec.Matches gets filtered out from every unwanted rules.

Fixes #747 

### How to test it

* `make test`
* Install traefik mesh with acl=true, apply a TrafficTarget with and without an HTTPRouteGroup.

### DIFF
--- a/integration/testdata/acl_enabled/http/6.traffic-target.yaml
+++ b/integration/testdata/acl_enabled/http/6.traffic-target.yaml
@@ -1,3 +1,13 @@
+apiVersion: specs.smi-spec.io/v1alpha3
+kind: HTTPRouteGroup
+metadata:
+  name: all
+  namespace: test
+spec:
+  matches:
+    - name: all
+      methods: ["*"]
+
 ---
 apiVersion: access.smi-spec.io/v1alpha2
 kind: TrafficTarget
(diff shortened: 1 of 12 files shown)


kevinpollet (approved):
LGTM 👍

PR #772, opened 2020-10-27
Add DNS resolver and server

### LINKED ISSUE #746: Shadow service name length limit
### What did you do?

Create a service with a name and a namespace which once concatenated are closed to but less than 63 characters.

I expected to see a shadow service associated with this service.

### What did you see instead?

An error complaining about the shadow service being longer than 63 characters. When this situation happen we are also leaking ports (both udp and tcp port mappers)

### Output of controller log: (_What version of Traefik Mesh are you using?_)

```
time="2020-09-11T15:33:30Z" level=error msg="Unable to complete work...must be no more than 63 characters"
```

Version: 1.4

kevinpollet:
## What does this PR do?

This PR:

- Adds DNS resolver and server to not rely anymore on the `CoreDNS` rewrite plugin.
- Removes the `prepare` cmd because SMI check versions and the DNS patching will be done by the controller and DNS.
- Changes the default `namespace` configuration in commands to `default`.

Fixes #746

### How to test it

* `make test`
* `make test-integration`


jspdown (approved):
LGTM :+1:

tardich:
Any idea when this will be pushed to a new version?  Most of my services can't currently be shadowed, which makes my mesh installation rather useless

PR #801, opened 2021-09-21
Fix TCPRoute and TrafficTarget config in ACL example

kevinpollet:
## What does this PR do?

This PR fixes the non-working `TCPRoute` and `TrafficTarget` in the ACL example.


SantoDE (approved):
Ah, good catch 👀

PR #822, opened 2022-08-01
Test support for arm64

odidev:
## What does this PR do?

This PR adds **ARM64** job to **SemaphoreCI**. 
Modified files `mesh/.semaphore/semaphore.yml `for running the test for arm64 platform.

**Signed-off-by**: odidev <odidev@puresoftware.com>


ldez:
Hello,

We don't need to run all the tests on arm64 because we use nothing related to a specific arch.
Spending energy for that, through our CI, is not environment friendly and it's useless.

Also, the workflow already creates an image for the different arch.

PR #823, opened 2022-08-02
Fix CoreDNS Detection

skwair:
## What does this PR do?

This PR updates the CoreDNS detection. Previously we were only looking for the exact deployment name `coredns` in the `kube-system` namespace. The issue was that some Kubernetes distributions (RKE2 for example), use a different name for the CoreDNS deployment.

With this PR, we now look for a deployment having the label `kubernetes.io/name=CoreDNS`, which seems to be the case for most Kubernetes distributions. We still fall back to the old behavior if no deployment is found using the label because `kubeadm` does not add it for example (but it uses the default deployment name).

### How to test it

Unit tests have been updated to cover the new use-case.

## Additional notes

If we face more edge-cases like this one, we might want to consider adding a flag so users can directly set the name (and type?) of there DNS provider.


kevinpollet (approved):
LGTM

PR #824, opened 2022-08-02
Remove Maxium CoreDNS Version Requirement

### LINKED ISSUE #821: Please Support CoreDNS version 1.9.0
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/mesh/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik Mesh community forum](https://community.traefik.io/c/traefik-mesh) and didn't find any.

### Proposal

image

Please support CoreDNS version 1.9.0

### Background

I am using MicroK8s, and latest version of MicroK8s using CoreDNS 1.9.0 then i install traefik-mesh using helm chart and error message showing that "Unsuported CoreDNS version 1.9.0"

### Workarounds

I don't think there's any solution for MicroK8S, or installing different CoreDNS like downgrade the version, because CoreDNS in Microk8S has special configuration and it's included as Core Addons.

skwair:
## What does this PR do?

This PR proposes to remove the artificial maximum CoreDNS version limit.

Most new CoreDNS versions are not breaking, but since we enforce a maximum version, we still need a PR to bump it for each new minor release.

With this PR, we assume it's going to work just fine (which should be the case more often than not), and if it doesn't, then we'll make a PR to fix it.

Fixes #821

### How to test it

Mesh should now work in Kubernetes environments with CoreDNS >= v1.9.0.


kevinpollet (approved):
LGTM

PR #832, opened 2022-09-23
fix code annotation  and update deps

nisainan:
## What does this PR do?

This PR:

fix code annotation  and update deps, replace value receiver with  pointer receiver  of struct Provider

### How to test it

No test required

## Additional Notes

Struct Provider has methods on both value and pointer receivers. Such usage is not recommended by the Go Documentation


ldez:
Hello,

A PR with several subjects is not a good practice, you will have to split your PR to create one PR by subject.

You will also have to explain your changes (ex: why did you update each dependency)

PR #878, opened 2026-07-08
chore(security): run image as non-root and disable CI checkout credential persistence

darkweaver87:
# Motivation

The runtime images already define a non-root `app` user (uid 1000) but never switch to it, so the entrypoint runs as root. Separately, the `actions/checkout` steps persist the GitHub token in `.git/config` for the whole job (CWE-522/200). This adds `USER app` to both Dockerfiles and `persist-credentials: false` to the two workflow checkouts (neither runs authenticated git afterwards).

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -22,6 +22,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
+          persist-credentials: false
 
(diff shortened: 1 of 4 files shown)


jspdown (approved):

LBF38 (approved):

darkweaver87:
Superseded by #879 (same change; branch/description cleaned up).

